PR #85, opened 2011-07-11
Add support for Tempita templates

mgorven:
Adds a `render_tempita` class in `web.contrib.templates` to allow [Tempita](http://pythonpaste.org/tempita/) templates to be used.


iredmail:
Feel free to reopen if it's still necessary.

PR #86, opened 2011-07-11
Add optional support for sessions without client cookies

mgorven:
If enabled and the client hasn't sent a session cookie, injects the session ID in the response by adding hidden form fields and adding query parameters to links. Requires lxml to process (X)HTML.


iredmail:
No plan for this feature recently. Feel free to reopen this issue if necessary.

PR #113, opened 2011-10-31
added support for X-HTTP-method-override header.  only tested running und

notatoad:
added support for X-HTTP-method-override header.  only tested running under isapi/iis.  i'm very noob, and have no idea if i've done this right or not, but it works for my needs.  i'd love to hear anybody's feedback.


iredmail:
Issue closed until someone asking for this feature.

PR #174, opened 2012-08-15
Please consider the MemoryStore class by Michele Petrazzo, thanks

ghost:
Store sessions in memory instead of database or file.
I'm using this class since some years in production without any problems in an embedded system.


iredmail:
Please add required unittests.

PR #193, opened 2012-12-02
Session 'initializer': Fixed Callable case of _initializer() in web.session.Session _load() 

mekarpeles:
in web.session.Session _load()

TLDR; Fixed _initialize for callable case. If initializer kwarg present, allow _initializer() to return dict which can be used to update self, i.e. for Session (assuming _initializer returns a dict) self.update(deepcopy(_initializer())) 

Long explanation:
Updated Session._load(self), within the logic for self._initializer (determining if _initializer is present, and if so whether it is a dict versus callable func). If _initializer is callable, right now there is no way to update the Session's attributes (like one could do if they just provide initialize=some_dictionary). One bad solution is to require _initializer take an argument of self. The better solution (implemented here) is for _load() to allow/expect the callable case of _initializer to return a dictionary which can be used to update the Session's attributes, i.e. self.update(_initializer())

### DIFF
--- a/web/session.py
+++ b/web/session.py
@@ -114,7 +114,9 @@ class Session(object):
                 if isinstance(self._initializer, dict):
                     self.update(deepcopy(self._initializer))
                 elif hasattr(self._initializer, '__call__'):
-                    self._initializer()
+                    attrs = self._initializer()
+                    if attrs and type(attrs) is dict:
+                        self.update(deepcopy(attrs))
  
         self.ip = web.ctx.ip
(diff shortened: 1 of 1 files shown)


iredmail:
Hi @mekarpeles 

- Are you still running web.py in production and need this feature?
- I usualy run code like `session['logged'] = True` to update session data, i personally don't think this suggested feature is better.

mekarpeles:
I think this can be closed.

PR #206, opened 2013-01-14
Added support for 'fdb' Firebird driver, fixed a sequence-related insert issue

benthepoet:
Added a line to try and import the new 'fdb' driver first, if it's not found then fall back to 'kinterbasdb'. 

Discovered in the db.insert there was an issue with the current fdb driver (0.9.9). Calling fetchone() after the insert produces a segmentation fault. Possibly because by default there's nothing for Firebird to return. Added a select query that will return a single column with a NULL value to prevent this from occurring.

Also found there was no logic for returning sequence values when doing an insert with FirebirdDB. Added the appropriate _process_insert_query to handle this. In order for Firebird to return a sequence value the sequence name must be provided. There's no definitive way to know what a table's sequence is in Firebird because auto incrementing fields are implemented using generators and triggers.

### DIFF
--- a/web/db.py
+++ b/web/db.py
@@ -1037,11 +1037,8 @@ class FirebirdDB(DB):
     """Firebird Database.
     """
     def __init__(self, **keywords):
-        try:
-            import kinterbasdb as db
-        except Exception:
-            db = None
-            pass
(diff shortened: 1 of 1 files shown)


iredmail:
Is this still required?

Issue #214, opened 2013-03-11
re_compile is not thread-safe (as the comment indicates) and causes "error: release unlocked lock" errors under load when used by application.py's context mapping


daubman:
As hinted at in the comment in:
https://github.com/webpy/webpy/blob/master/web/utils.py#L509
"re_compile = memoize(re.compile) #@@ threadsafe?"

The memoization of re.compile does not appear to be threadsafe and is causing:
"error: release unlocked lock"
errors under load when used by application.py's context mapping code at:
https://github.com/webpy/webpy/blob/master/web/application.py#L229

A full stack trace follows:
"""
Traceback (most recent call last):
  File "/var/www/worker_code/main.py", line 1909, in handle
    return web.application.handle(self)
  File "/opt/current-env/lib/python2.6/site-packages/web/application.py", line 229, in handle
    fn, args = self._match(self.mapping, web.ctx.path)
  File "/opt/current-env/lib/python2.6/site-packages/web/application.py", line 435, in _match
    what, result = utils.re_subm('^' + pat + '$', what, value)
  File "/opt/current-env/lib/python2.6/site-packages/web/utils.py", line 531, in re_subm
    compiled_pat = re_compile(pat)
  File "/opt/current-env/lib/python2.6/site-packages/web/utils.py", line 499, in __call__
    update(block=True)
  File "/opt/current-env/lib/python2.6/site-packages/web/utils.py", line 496, in update
    self.running[key].release()
error: release unlocked lock
"""

PR #215, opened 2013-03-12
allow the optional use of a path-to-class dict instead of regex-matching all paths, workaround Issue #214 and general perf improvement

daubman:
If you have a constant (no regex matching necessary) set of paths to map, you can now add use_path_dict=True to your:
app = web.application(urls, globals(), use_path_dict=True)
call to use a dict of something like:
{'web.ctx.path-1': class_to_handle_path-1}

rather than avoiding the regex matching against all paths for every call.

While this will have a tiny perf boost, its main benefit is avoiding the non-thread-safe memoization of re_compile discussed in Issue #214

This should be completely transparent and backwards compatible for any users requiring regex matching of paths...

### DIFF
--- a/web/application.py
+++ b/web/application.py
@@ -110,7 +110,8 @@ class application:
     def _cleanup(self):
         # Threads can be recycled by WSGI servers.
         # Clearing up all thread-local state to avoid interefereing with subsequent requests.
-        utils.ThreadedDict.clear_all()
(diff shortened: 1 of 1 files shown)


iredmail:
No plan to implement this, feel free to reopen the this issue if required. 

Btw, I personally don’t have any experience with an app which doesn’t need URL regex, MAYBE too rare and not worth to implement it.

PR #218, opened 2013-03-28
added a method to renew session id on auth to prevent session fixation

vogan:
call this method during user authentication to prevent session fixation
attacks. the old, unauthenticated session id is re-generated in order to
start a "new" session after login


iredmail:
Do we really need this feature?

iredmail:
Closed due to no plan to implement it. Feel free to reopen the issue if required.

PR #223, opened 2013-04-15
New Store mechanism to use with Google App Engine

eduardohenriquearnold:
Creates a Store mechanism to use with sessions when developing to Google App Engine. This is based on their own data storage system. Other methods based on file storage don't work because GAE doesn't allow file writing.
Credits to yejun (https://github.com/yejun)

An example code is:

``` python
session = web.session.Session(app, web.session.GoogleStore())
```

### DIFF
--- a/web/session.py
+++ b/web/session.py
@@ -352,6 +352,45 @@ class ShelfStore:
             atime, v = self.shelf[k]
             if now - atime > timeout :
                 del self[k]
+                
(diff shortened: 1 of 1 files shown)


iredmail:
Dear @eduardohenriquearnold 

Are you still running web.py on Google App Engine? is this feature still necessary?

Issue #182, opened 2012-09-24
Race condition in setitem() of session.DiskStore


pbuckner:
DiskStore.**setitem**() does f.open(), f.write(), f.close() of session data. Problem is, if different thread attempt to open and read the same file, while the file is already opened for f.write(). The read returns zero bytes, resulting in session data decode failure, which percolates back as a session failure.

Solution can be to instead open a temporary file, write & close the temporary file, and then rename it to the proper file. This is guaranteed atomic. That way, there's always a valid session DiskStore.

session.py, approx line 260:

```
class DiskStore(Store):
....
  def __setitem__(self, key, value):
    path = ...
    pickled = ...
    try:
        f = tempfile.NamedTemporaryFile(delete=False)  # use tempfile rather that path
        try:
            f.write(pickled)
        finally:
            f.close()
            os.rename(f.name, path)  # update path atomically
    except IOError:
        pass
```

Issue #83, opened 2011-07-07
EOFError when loading empty session files


leres:
It appears for whatever reason I occasionally end up empty web.py session files:

<pre>
File "/usr/local/lib/python2.7/site-packages/web/session.py", line 199, in decode, referer: https://onestop.lbl.gov/onestop/
    return pickle.loads(pickled), referer: https://xyz.lbl.gov/webapp/
EOFError, referer: https://xyz.lbl.gov/webapp/
</pre>

It seems like web.py could catch the EOFError and pretend there was no session file. If that isn't right, should I catch the call to web.session.Session()?

<pre>
if web.config.get('_session') is None:
    session = web.session.Session(APP,
        web.session.DiskStore(os.path.join(ROOT, 'sessions')), {'user': 'anonymous'})
    web.config._session = session
else:
    session = web.config._session
</pre>

What's the preferred way to deal with this?

PR #224, opened 2013-04-16
Fixed #191, #182, #83: DiskStore concurrency issue

### LINKED ISSUE #191: session.DiskStore concurrency issue

<pre>
    path = self._get_path(key)
    pickled = self.encode(value)
    try:
        f = open(tname, 'w')
        try:
            f.write(pickled)
        finally:
            f.close()
    except IOError:
        pass
</pre>

HansWeltar:
See #191, #182, #83
DiskStore.**setitem** is not atomary and a host of errors can be thrown under
load. This method has been made threadsafe.

### DIFF
--- a/web/session.py
+++ b/web/session.py
@@ -5,6 +5,7 @@ Session Management
 
 import os, time, datetime, random, base64
 import os.path
+import threading
 from copy import deepcopy
 try:
     import cPickle as pickle


iredmail:
Fixed in https://github.com/webpy/webpy/commit/058672df9cf7b717da7b0f5d0540875853940862

PR #232, opened 2013-06-04
Set default cookie_path to '/'

dieyushi:
When deploying with mod_wsgi, every request will generate a new session by default. The session sample in the cookbook dosn't work at all. An optional solution is set 'web.config.session_parameters['cookie_path']' before using session. Set default cookie_path will simply solve this issue.


blueyed:
I also think that the documentation/cookbook should be adjusted instead.

PR #238, opened 2013-07-18
Matching unicode in URL patterns.

toyg:
web.py is very clever at translating unicode urls, but then uses ascii-only regexes to (not) route them. These little changes in application.py and utils.py allow proper matching of unicode urls.

### DIFF
--- a/web/application.py
+++ b/web/application.py
@@ -2,6 +2,7 @@
 Web application
 (from web.py)
 """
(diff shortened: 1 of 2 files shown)


iredmail:
hi @toyg 

- Do you still run web.py application in production?
- Is this PR still necessary?
- If necessary, could you please add required unittests in this PR?

PR #239, opened 2013-07-26
Fix DiskStore concurrency issue.

### LINKED ISSUE #191: session.DiskStore concurrency issue

skawouter:
DiskStore.**setitem** is not atomary and a host of errors can be thrown under
load. This method has been made threadsafe.

This fixes #191, #182, #83

### DIFF
--- a/web/form.py
+++ b/web/form.py
@@ -152,6 +152,12 @@ class Input(object):
 
     See also: <https://www.w3.org/TR/html52/sec-forms.html#the-input-element>
 
+    Currently only types which can be written inside one `<input />` tag are
+    supported.
+
+    - For checkbox, please use `Checkbox` class for better control.
+    - For radiobox, please use `Radio` class for better control.
+
         >>> Input(name='foo', type='email', value="user@domain.com").render()
         '<input id="foo" name="foo" type="email" value="user@domain.com"/>'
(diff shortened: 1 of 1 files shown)


iredmail:
Hi @skawouter @jzellman 

- Are you still run web.py application in production?
- Is this PR still necessary?

skawouter:
> Hi @skawouter @jzellman
> 
> * Are you still run web.py application in production?
> * Is this PR still necessary?

Hi @iredmail 

Yes we still run this in production with this patch applied. 
I do think this is still needed if you run web.py threaded with a lot of load the file could still become corrupted.

iredmail:
Hi @skawoute

Could you generate a new patch based on current git version of web.py? I will merge it later. 

Btw, any plan to move your app to py3? We fixed lots of py3 compatibility issues recently, it would be very helpful if you can try to run your app with latest web.py + python3, report the web.py issues so that we can fix it and deliver a new stable release. :)

skawouter:
@iredmail I created a new pull request #537 that can be applied to master
We will begin testing with Python3 soon. And will report and or fix any issues we might find.

PR #245, opened 2013-08-28
Python test

mugongjian:
i fix test/db.py to support python3


iredmail:
@cclauss Anand is not active, i suggest we handle this manually instead.

iredmail:
Changes in this PR have been implemented in latest git version.

PR #255, opened 2013-10-31
Fixed error in creating pooled PostgresDB with pgdb driver.

PriceChild:
Repeating the same fix in 35674c8cd49e29853e8728fdf5fdb4ab0da5b58a for pooled (With DBUtils) connections.

### DIFF
--- a/web/db.py
+++ b/web/db.py
@@ -970,7 +970,11 @@ class PostgresDB(DB):
         
     def _connect_with_pooling(self, keywords):
         conn = DB._connect_with_pooling(self, keywords)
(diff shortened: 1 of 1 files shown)


PriceChild:
@iredmail I believe so, yes.

I don't have the same environment around 6 years later, but reviewing the code quickly it looks like connections without DBUtils still fall back: https://github.com/webpy/webpy/blob/master/web/db.py#L1037 but those with pooling don't: https://github.com/webpy/webpy/blob/master/web/db.py#L1048

PR #263, opened 2013-11-13
fix multiple_insert in mysql return insert ids not right reference 

mugongjian:
i post the issue on wepy groups,https://groups.google.com/forum/#!topic/webpy/KLM95_M5SDw.


nlaurens:
I ran into the same issue. Proposed fix works for mysql (but breaks other dbs). Here is a proposed fix that does not alter behaviour for other dbs: #446

PR #280, opened 2014-02-19
thousands_separator and decimal_point in web.utils.commify

nopri:
Old behavior: 
We can't specify decimal point or thousands separator in web.utils.commify()

New behavior:
Both decimal point and thousands separator can be optionally set.
Note:
- Default value for decimal point (decimal_point) is '.'
- Default value for thousands separator (thousands_separator) is ','

Added example:

``` python
>>> commify(1234.5, decimal_point=',', thousands_separator='.')
'1.234,5'
>>> commify('%.2f' %1234.5, decimal_point=',', thousands_separator='.')
'1.234,50'        
```

### DIFF
--- a/web/utils.py
+++ b/web/utils.py
@@ -966,7 +966,7 @@ def denumify(string, pattern):
             out.append(c)
     return ''.join(out)
 
-def commify(n):
+def commify(n, decimal_point='.', thousands_separator=','):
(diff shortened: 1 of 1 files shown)


iredmail:
No plan for this change. Sorry.

PR #284, opened 2014-03-12
modernizing Templetor SafeVisitor code using new ast module

egbutter:
I noticed that some of the newer dictionary comprehensions were getting denied by SafeVisitor() .. digging in, this was b/c they were not in the allowed node WHITELIST.  Since we are only looking for a few very recognizable types of "unsafe" nodes, it seems more maintainable to make this a BLACKLIST.  The new ast module also provides a NodeVisitor class now, which allows us to clean up this SafeVisitor class a little bit, and make it Py3 compatable.  Also, the full ast module is available in jython: http://www.jython.org/docs/library/ast.html.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.pyc
 .DS_Store
+web.py.egg-info
--- a/web/net.py
+++ b/web/net.py
@@ -19,6 +19,8 @@ import socket
 def validip6addr(address):
     """
     Returns True if `address` is a valid IPv6 address.
+    catch AttributeError - python does not handle ipv6 errors normally
+    https://github.com/webpy/webpy/pull/195
 
         >>> validip6addr('::')
         True
@@ -31,7 +33,7 @@ def validip6addr(address):
     """
     try:
         socket.inet_pton(socket.AF_INET6, address)
-    except socket.error:
+    except (socket.error, AttributeError):
         return False
 
     return True
--- a/web/template.py
+++ b/web/template.py
@@ -1151,18 +1151,21 @@ class SafeVisitor(ast.NodeVisitor):
         super(SafeVisitor, self).generic_visit(node)
 
     def visit_Assign(self, node):
-        self.check_assign_node(node)
+        self.check_assign_targets(node)
 
     def visit_AugAssign(self, node):
-        self.check_assign_node(node)
+        self.check_assign_target(node)
 
-    def check_assign_node(self, node):
+    def check_assign_targets(self, node):
         for target in node.targets:
-            targetname = type(target).__name__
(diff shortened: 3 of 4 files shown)


jzellman:
I like the idea of using the ast library. Is it available under all versions of python web.py supports? Also there are some changes (e.g. docstrings) that are outside of this change. It would be much easier to read/merge if the changes were limited to what the task is, modernizing SafeVisitor.  Coming up on 3 years, but anyway you can reduce this PR to the task at hand? And then open separate PRs for the other changes?

egbutter:
hi @jzellman ast is not available until v2.6. web.py used to support even older versions, I'm not sure if that's still your policy, but by now i imagine people have upgraded to higher 2x versions or python 3. Supporting both at the same time is going to be ugly, though totally doable. :-1: 

I'd be happy to split this out into three PR's for you and resubmit. 1. docstrings on the old code, 2. ast upgrade, 3. whitelisting upgrade.

Before I do so: there's not an exact 1:1 mapping between the old and new tree, so this does introduce new risks either way. I submitted the two items together for a reason: 1. web.py is very stable but has limited community support, so fixes are introduced very slowly (as evidenced by the PR backlog), while 2. there are regular upgrades to the language (esp py3) that are going to get missed with the current blacklist technique, opening up the templating engine to unnecessary exploits.

jzellman:
Thanks @egbutter. Looking at the separate commits, they seem to address individual things. I'm OK with leaving it as 1 PR with individual commits. Looking at .travis.yml (https://github.com/webpy/webpy/blob/master/.travis.yml), the project supports 2.6 and 2.7 which appear to have ast. I am not the maintainer of the project and I don't know what the consensus is to drop v2.x support, but I know personally I will be disappointed. I still use the latest 2.7 version. 

Unfortunately though, it looks like v2.6 and v2.7 ci builds fail. 1 is due to the date changes. Is this potentially an incompatibility between Python 2 and 3? The other is due to dict comprehensions not being supported in v2.6. Here's a link to the failed builds: https://travis-ci.org/webpy/webpy/jobs/59916978 https://travis-ci.org/webpy/webpy/jobs/59916979

The lack of dict comprehension can be written in a less elegant way using tuples and dict constructor:
```
>>> nums = range(10)
>>> import string
>>> chars = string.ascii_letters[:10]
>>> dict([(c, n) for c, n in zip(chars, nums)])
{'a': 0, 'c': 2, 'b': 1, 'e': 4, 'd': 3, 'g': 6, 'f': 5, 'i': 8, 'h': 7, 'j': 9}
```

cclauss:
Conflicts: Please rebase this PR.  Cool stuff!

PR #285, opened 2014-04-07
Auth module for webpy

phoracek:
Hello, I created module which handles authorization and authentication in webpy applications. Could you consider it as a part of the framework? I think that this is an important and handy feature.

Module itself is really easy to use. It uses SHA256 and Bcrypt crypting algorithms as default, but there is no problem to extend it. It also contains unit test and documentation, where you can learn more about it.

This is my first real contribution to open-source project, I'll accept all critics about the code, semantics etc.

Basic example (not all functions):

```
db = web.database(...)
session = web.session.Session(...)
auth = web.auth.Auth(session, db, lgn_pg='/login')

auth.login(usr, passwd)

@auth.role('admin')
def GET(self):
    return render.text("Admin's page")

auth.logout()
```

### DIFF
--- a/web/auth.py
+++ b/web/auth.py
@@ -108,7 +108,8 @@ class Auth(object):
             user = self._ses.usr
             row = self._getrow(user)
             return row.role
-        except AttributeError:
+        except AttributeError as e:
+            print "AttributeError:", e
             return None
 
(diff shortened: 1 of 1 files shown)


jzellman:
@anandology could you please close?

PR #290, opened 2014-05-03
Implemented CSRF token for forms and changed rendernote usage

SmileyJames2:
I have implemented a CSRF token form input.
Please can someone carefully check that this implementation is secure.
I have also changed the implementation of rendernote, so that Form calls an inputs own rendernote method, so that you can override the rendernote method if needed.


jzellman:
I think this would be better suited as an item in the cookbook or in a separate library. Something like "web.py extras". @anandology thoughts on this?

PR #292, opened 2014-05-12
Python 2.5 no longer supported on Travis CI

frewsxcv:
Which will fix the broken build status indicator


charmander:
This has now been addressed in 7f629f4.

PR #310, opened 2014-11-06
Make re-compile thread-safe (Issue #214)

HansWeltar:
Under heavy load, we sometimes see the following error (see issue #214)
Exception in thread Thread-328:
  File "/pkg/lib/python2.7/site-packages/web/utils.py", line 500, in **call**
    update(block=True)
  File "/pkg/lib/python2.7/site-packages/web/utils.py", line 497, in update
    self.running[key].release()
error: release unlocked lock

This can happen when there is a thread-switch between the test and creation
of self.running[key]. In this patch we acquire a lock first
to make the test+create atomic.

### DIFF
--- a/web/utils.py
+++ b/web/utils.py
@@ -483,11 +483,13 @@ class Memoize:
         self.expires = expires
         self.background = background
         self.running = {}
+        self.running_lock = threading.Lock()
     
     def __call__(self, *args, **keywords):
         key = (args, tuple(keywords.items()))
-        if not self.running.get(key):
-            self.running[key] = threading.Lock()
+        with self.running_lock:
+            if not self.running.get(key):
+                self.running[key] = threading.Lock()
         def update(block=False):
             if self.running[key].acquire(block):
                 try:
@@ -506,7 +508,7 @@ class Memoize:
 
 memoize = Memoize
 
-re_compile = memoize(re.compile) #@@ threadsafe?
+re_compile = memoize(re.compile)
 re_compile.__doc__ = """
 A memoized version of re.compile.
 """


Rimco:
Even with low load, but with simultaneous AJAX requests this error occurs. I agree with this solution and would like to see it integrated.

jzellman:
Does this happen with the built in dev server or when sitting behind something like apache, nginx etc?

Rimco:
To me it happened as part of https://github.com/Rimco/OSPy/tree/refactor, without any other webserver in between.

jzellman:
Thanks @Rimco and applying the above change set resolves the issue?

Rimco:
It's too long ago to say for sure that I tried the fix. At least I remember I had exactly the same signature and it was happening during simultaneous AJAX requests. If I look at the fix, it seems to be a logical fix at least.

jzellman:
@HansWeltar things like this are nearly impossible (and probably not worth the effort) to build a continuous test around. Can you provide some details as to how you verified this worked, environment, conditions, etc?

HansWeltar:
We run web.py behind lighttpd on a linux-based system.
You are correct that we could never reproduce it reliably.
We usually saw the error when we did many simultaneous AJAX requests.
Before applying the fix we saw the error regularly.
Since we applied this fix we haven't seen the error since.

I did not create a unittest for this particular case. However i think you could spawn 2 threads that both invoke __call__ a bunch of times. This still does not guarantee that the error will show, but in practice i've had good success with this technique and about 1000 iterations in 2 threads.

HansWeltar:
The code below seems to trigger it mostly (test is still not 100% reliable but running it several times you should be able to trigger it at least once).
With the patch in place i'm no longer able to trigger the error.

```python
import web
import threading
import time

m=web.utils.Memoize(lambda x: x+1)

def loop1():
  for i in range(1000000):
    m(i)  # this code is not threadsafe

# spawn two threads and hope they trigger the non-threadsafe behaviour
t1 = threading.Thread(target=loop1)
t2 = threading.Thread(target=loop1)
t1.start()
t2.start()
t1.join()
t2.join()
```

anandology:
Thanks @HansWeltar. This is very helpful.

jzellman:
Tested using @HansWeltar's provided code. Was able to repeatedly hit the threading issue. After applying his change, I could not recreate it. 👍 :shipit:

Gert-Jan4C:
Great that this thing is fixed; I came across this issue when investigating some test case instabilities. However, I don't see it is resolved in any published version of webpy? I am using v0.38 now and it has the threading problem.

Will this fix be released to an official version soon?

iredmail:
Committed the changes in pull request.

PR #313, opened 2014-11-22
solved python3 incompatible syntax issues that popped up during setup install of python3 branch

proycon:
..mostly exception syntax


anandology:
All the python 3 syntax issues have been fixed the master now.

PR #314, opened 2014-11-25
Fix data loss issue with transaction error handling

benhoyt:
If a database error (for example a connection error) occurs during the `self.engine.do_commit()` or `self.engine.do_rollback()` calls here, then the transaction won't be popped off the transaction stack.

Note that I'm using web.py with connection pooling (DBUtils), which are good for fast, long-running connections. So what was happening was all the subsequent database transactions would _appear_ to succeed, but actually were never committed because there was still something on the transaction stack at the end of the next commit.

This fix corrects the issue. An alternative (I think) would be to simply swap the order of the two lines, ensuring the `self.ctx.transactions` update happens first and hence always happens.

### DIFF
--- a/web/db.py
+++ b/web/db.py
@@ -471,13 +471,17 @@ class Transaction:
 
     def commit(self):
         if len(self.ctx.transactions) > self.transaction_count:
-            self.engine.do_commit()
-            self.ctx.transactions = self.ctx.transactions[:self.transaction_count]
+            try:
+                self.engine.do_commit()
+            finally:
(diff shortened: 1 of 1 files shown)


iredmail:
I got exactly same output before and after patched:

```
Success: 0 157
Error: 1 InterfaceError('connection already closed',)
Success: 2 159
Success: 3 160

In database (same connection): 157
In database (same connection): 159
In database (same connection): 160
---------------
In database (new connection): 157
In database (new connection): 159
In database (new connection): 160
```

PR #317, opened 2015-01-24
Add support for xml templates to render_jinja2

iamFIREcracker:


jzellman:
Could you remove any extraneous white space changes and add to a simple test case? What happens if you have 2 files index.html and index.xml for example?

iamFIREcracker:
It's been awhile since I opened this, and TBH I don't remember what is gonna happen when you have both index.html/index.xml files around.  I don't even remember why I created this (it might have had something to do with dynamic sitemap.xml files), but I think the framework should favor html files over xml ones.

Also, where would you like me to put tests for this?

iredmail:
No plan to support this. sorry.

PR #319, opened 2015-02-09
Add support of SSL certificate chain.

movb:
According to http://webpy.org/cookbook/ssl you can provide only
certificate file and key file to ssl server in web.py.

Now you can also setup certificate chain file. For example:
    CherryPyWSGIServer.ssl_certificate = "/path/to/ssl_certificate"
    CherryPyWSGIServer.ssl_private_key = "/path/to/ssl_private_key"
    CherryPyWSGIServer.ssl_certificate_chain = "/path/to/ssl_certificate_chain"

### DIFF
--- a/web/httpserver.py
+++ b/web/httpserver.py
@@ -174,7 +174,7 @@ def WSGIServer(server_address, wsgi_app):
     
     server = wsgiserver.CherryPyWSGIServer(server_address, wsgi_app, server_name="localhost")
         
-    def create_ssl_adapter(cert, key):
+    def create_ssl_adapter(cert, key, chain=None):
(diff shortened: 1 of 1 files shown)


WinPack:
Still not integrated?

Tibabal:
For other needing that patch, install web.py using my fork applying this patch:

git clone https://github.com/Tibabal/webpy.git
ln -s `pwd`/webpy/web .

jzellman:
#389 appears to be same? :ship:

anandology:
CherryPy Webserver has been replaced with Cheroot. Please see http://webpy.org/cookbook/ssl for new instructions.

PR #320, opened 2015-02-14
Add redis-based session store

iamFIREcracker:
With this commit I would like to introduce a new session store based on Redis.  This store creates a sorted set with key `webpy.sessions` containing all the id of all the active sessions, and a key with name`webpy.session:$ID` containing the data associated with the active session.

### DIFF
--- a/web/session.py
+++ b/web/session.py
@@ -353,6 +353,54 @@ class ShelfStore:
             if now - atime > timeout :
                 del self[k]
 
+class RedisStore(Store):
+    """Store for saving a session in Redis.
+
+    This store will create a key `webpy.sessions` containing all the
+    active sessions, and a key for each of those sessions."""
+
+    def __init__(self, redis):
+        self._redis = redis
+
+    def _redis_key(self, key):
+        return 'webpy.session:' + key
+
+    def _datetime_to_float(self, dt):
+        return int(float(dt.strftime('%s.%f')) * 1000000)
+
+    def __contains__(self, key):
+        return self._redis.get(self._redis_key(key)) is not None
+
+    def __getitem__(self, key):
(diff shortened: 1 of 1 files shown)


jzellman:
I like the idea and could see myself using something like this. I do have a couple concerns with this PR though. web.py is meant to be small and simple. This adds a dependency/complexity to web.py on a particular redis library that supports things like zrem(?), etc. This PR also opens the door to supporting any other session store, like amazon S3, berkley db, memcache. The build also fails with no provided tests. 

Maybe this could be better suited in the cookbook as a sample of how to build your own custom SessionStore? Or maybe it belongs in the experimental or contrib directory? Another idea would be to have a separate project, say web.py extras where things like this could live. Thoughts?

iamFIREcracker:
I don't have a strong opinion on this, so just tell me what do you prefer me to do (eg. convert this into a cookbook example, move it somewhere inside experimental/contrib).

Once upon a time I created a repository[0] with a bunch of web.py related things that I constantly found myself in need of when starting a new project; if you find some of those things interesting I could create small PRs and try to get them merged -- but again, we might want to find a new home for all this stuff first (be it the experimentarl/contrib folder, or a new official repo)

[0] https://bitbucket.org/iamFIREcracker/weblib/src/7d0975cec10e?at=default

dmmartins:
I agree with the cookbook redis example or a generic example on how to write your own session store. I think you should put your work in a repo, though. Using redis as session store is very common and more people can use that. Just maybe not as part of core web.py.

Also as a suggestion, the 'webpy.session:*' key can be a default value but you should be able to use a different value.

BTW you repo weblib is not public.

iredmail:
No plan to implement Redis based session store yet, feel free to make it as a third-party repo for further use.

PR #322, opened 2015-03-07
db.rst: Mention how to close sql connection explicitly. (closed #251)

### LINKED ISSUE #251: Automatic Database Closing
I could not find in the source code anywhere that the database connection is actually closed. At least when using a mysql database, this results in aborted clients. 

In the DB class, if you attempt to call ctx.db.close() if close is an available function, it cleanly closes the db connection and releases those resources. 

Please think about adding code similar to :

<code>
class DB:
 def **del**(self):
  ctx.close()
</code>

zhanghuangbin:
Mention how to close sql connection explicitly. (closed #251)


jzellman:
@anandology I'm not sure if it is possible, but is there a way for you to force travis ci to rebuild? Then I'd say :ship:

Issue #323, opened 2015-03-15
atime is not used for all database calls in session.py leading to unexplained failures


suhashpatil:
The web.py framework depends upon default value of atime being set by Mysql and hence not all the database calls that create or update session variable have atime as field, update call has it but insert does not. 
The file python2.7/site-packages/web/session.py has this code: 
        self.db.insert(self.table, False, session_id=key, data=pickled )
whereas update has this:
        self.db.update(self.table, where="session_id=$key", atime=now, vars=locals())

I use Mysql 5.5 which does not support default value for atime on datetime field but supports for timestamp field. The instructions for web.py are for 5.6 and above(apparently) and hence I could not set default value for atime and ignored the instructions not realizing that framework code depends upon this. If in such a case a session is created and another is created without causing an update to the first session then the first session variable is deleted from the database as the defeault value for that session is set Null. 

Simply adding "atime=now" to all db related calls will fix this problem. 

I ran into it because I wrote tests that depended upon me creating multiple sessions first and then using those sessions.

anandology:
Thanks @suhashpatil. The PR has been merged.

PR #324, opened 2015-03-15
fix for https://github.com/webpy/webpy/issues/323

suhashpatil:
fix for https://github.com/webpy/webpy/issues/323


anandology:
Thanks @suhashpatil.

PR #326, opened 2015-03-29
added ctx.db.close() method to class destructor ref issue: #251

micahschaefer:
per discussion in issue: #251

This is how I patched db.py to explicitly close database connections on class destruction; which solves the issue with MySQL abandoned clients while waiting for garbage collection.

I was using multi-threading and opening hundreds of database connections at a time, which would crash MYSQL after a couple rounds

### DIFF
--- a/web/db.py
+++ b/web/db.py
@@ -925,6 +925,10 @@ class DB:
     def transaction(self): 
         """Start a transaction."""
         return Transaction(self.ctx)
+
+    def __del__(self):
+        """faster close of database connection"""
+        self.ctx.db.close()
     
 class PostgresDB(DB): 
     """Postgres driver."""


anandology:
Your patch is only closing the connection from the current thread. What about the connections opened by all other threads?

kulous:
Each thread would have its own db instance, as you couldnt share across threads. So as a child process would close, a db instance was sitting not closing a connection causing an abandon

anandology:
The whole point of having the ctx is to make the db instance work for
multiple threads.


On Sun, Mar 29, 2015 at 9:22 AM, Micah Schaefer <notifications@github.com>
wrote:

> Each thread would have its own db instance, as you couldnt share across
> threads. So as a child process would close, a db instance was sitting not
> closing a connection causing an abandon
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/webpy/webpy/pull/326#issuecomment-87349299>.
>



-- 
Anand
http://anandology.com/

kulous:
I'm not sure how that works then. I haven't used this is a while, but I wanted to explain my issue for the other guy as best as I could. 

I can close out this PR and the issue if you deem fit?

jzellman:
@anandology can you close this for now?

PR #327, opened 2015-04-11
Provide support for multiple input fields

veprbl:
Most of the things were already working except for webapi.input() getting utils.storify() to filter out the duplicate keys. This will now allow to tell utils.storify() to not do that by simply setting form's default value to a list.

The feature can be tested using the following example:

``` python
import web
from web import form

urls = ('/', 'index')

myform = form.Form(
    form.Dropdown(name='car_type[]', multiple='multiple', args=[("a", "a"), ("b", "b")], value=["a", "b"]),
    form.Button("submit", type="submit"))

class index:
    def GET(self):
        form = myform()
        web.header('Content-Type', 'text/html')
        res = ""
        if form.validates():
            res += "car_type[] = %s" % repr(form["car_type[]"].value)
        res += """<form method="GET">%s</form>""" % form.render()
        return res

if __name__=="__main__":
    web.application(urls, globals()).run()
```

### DIFF
--- a/docs/input.rst
+++ b/docs/input.rst
@@ -60,6 +60,25 @@ What if you have a URL which looks like `/page?id=1&id=2&id=3` or you have a for
             ids = data.id         # now, `ids` is a list with all the `id`s.
             ...
 
+Same trick works for forms
+
+::
+
+    myform = form.Form(
+        form.Dropdown(name="id", multiple="multiple", args=[("a", "a"), ("b", "b")], value=[]),
+        form.Button("submit", type="submit")
+        )
+
+    class index:
+        def GET(self):
+            form = myform()
+            web.header("Content-Type", "text/html")
+            res = ""
+            if form.validates():
+                res += "id = {}".format(repr(form["id"].value))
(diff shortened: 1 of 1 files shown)


jzellman:
Could you please add a test case

veprbl:
@jzellman done

web/form.py
```diff
@@ -70,7 +70,8 @@ def rendernote(self, note):
         else: return ""
     
     def validates(self, source=None, _validate=True, **kw):
-        source = source or kw or web.input()
+        defaults = dict([(i.name, i.get_value()) for i in self.inputs])
```
jzellman: I don't believe this more subtle format is supported under 2.6?
```defaults = {i.name: i.get_value() for i in self.inputs}```

jzellman:
Other than the 1 comment (which I don't believe python 2.6 supports dict comprehensions), i'd say :ship:

web/form.py
veprbl: No, this is not supported.

PR #328, opened 2015-04-14
Reject route matches ending with newlines

charmander:
`$` allows a single line terminator before the end of the string, causing a KeyError when a route would otherwise match. This corrects the result to a “not found”.
#316 in a new branch.

### DIFF
--- a/test/application.py
+++ b/test/application.py
@@ -92,6 +92,20 @@ class ApplicationTest(webtest.TestCase):
         response = app.request('/b/foo?x=2')
         self.assertEquals(response.status, '301 Moved Permanently')
         self.assertEquals(response.headers['Location'], 'http://0.0.0.0:8080/hello/foo?x=2')
+
+    def test_routing(self):
+        urls = (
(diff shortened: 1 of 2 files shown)


test/application.py
```diff
@@ -92,6 +92,19 @@ def GET(self, name):
...
+            "/foo", "foo"
+        )
+
+        class foo:
+            def GET(self):
+                return "foo"
+
+        app = web.application(urls, {"foo": foo})
+
+        self.assertEquals(app.request('/foo\n').data, b'not found')
```
jzellman: Is there a test that covers the good case? e.g. self.assertEquals(app.request('/foo').data, 'foo') ?

jzellman:
Can you give me an example of this in the wild? Right now URL matching also ignores trailing whitespace.

test/application.py
charmander: Added.

anandology:
Thanks @charmander for the PR. @jzellman thanks for reviewing.

PR #333, opened 2015-06-15
some db.py error fix

bigzhu:
1. make sure keys are sorted, keep compare correct in multiple_insert
2. when seqname is not provided guess the seqname will cause error: 
   
   > psycopg2.OperationalError: currval of sequence "xxx" is not yet defined in this session
   
   when you insert id key in postgresql 9.4
   I'm disable the seqname guess, if you need insert return id, must provide seqname. It may cause some problems if you use seqname guess return insert id.

### DIFF
--- a/web/db.py
+++ b/web/db.py
@@ -811,7 +811,7 @@ class DB:
             >>> db.supports_multiple_insert = True
             >>> values = [{"name": "foo", "email": "foo@example.com"}, {"name": "bar", "email": "bar@example.com"}]
             >>> db.multiple_insert('person', values=values, _test=True)
-            <sql: "INSERT INTO person (name, email) VALUES ('foo', 'foo@example.com'), ('bar', 'bar@example.com')">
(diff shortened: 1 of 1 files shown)


jzellman:
After looking through the changes, I'm having a hard time seeing the differences since there are a number of diffs that are not part of this change like spacing changes. Could you please revert any erroneous changes to make the diff as small as possible. This will aid merging and will minimize the amount of time needed to review.

bigzhu:
Ok, I will repull this

bigzhu:
@jzellman This bug is fixed in current version, so my fix is no need.

anandology:
@bigzhu it is still an issue with Python 2. I'm able to successfully reproduce the issue. It is not an issue with Python 3 though. 

Here is code to reproduce it.

	import web

	db = web.database(dbn="postgres", db="rx-datasets")

	d1 = {"x": 1, "y": 2, "z": 3}

	d2 = {i:i for i in range(1000)}
	d2.update(d1)
	for i in range(1000): del d2[i]

	db.multiple_insert("test2", [d1, d2])

I'll look into this issue shortly.

PR #335, opened 2015-07-13
modify utils.py for mutiple file upload

SUNNYWILLHAPPEN:
can support multiple file upload by one submit.
of course.
The form needs an attribute name="mutiplefile", or this won't work correctly.
In the webpy code, a default value is needed (the "myfile["mutiplefile"] = []" part) if you want it to be imported as a CGI FieldStorage object. 
i donnot think this is a good solution,but it can works. or maybe you have a  more good way


jzellman:
After reviewing the changes, I'm having a hard time seeing the differences since there are a number of diffs that are not part of this change like spacing changes etc. Could you please revert any erroneous changes that are not part of the PR?

PR #337, opened 2015-07-24
make it possible to set expire time of cookies in browser.Before this…

jianqiao-ms:
…,the Expires of cookies will always be 'Session' which means that the cookie will be expired when the browser closed(or close tabs).In this situation,timeout of the session is meanless.Now,with the modify we can set a value of seconds in web (via select of other things) make cookie stored in clients for a period.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.pyc
 .DS_Store
+.idea/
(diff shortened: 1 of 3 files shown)


jzellman:
After looking through the changes, I'm having a hard time seeing the differences since there are a number of diffs that are not part of this change like spacing changes and changes to .gitignore. Could you please clean them up and squash all changes down to 1 commit?

PR #340, opened 2015-09-13
add interceptor support

quinnzhuang:
Add interceptor support for webpy


jzellman:
web.py already supports application processors. How does this to application processors?

anandology:
I agree. This is not very different from application processors. Rejecting this PR. 

@jzellman Thanks for reviewing this.

PR #344, opened 2015-10-08
Modify experimental/background.py to work with webpy 0.3x. Add an example.

LiberiFatali:
To initialize web._context properly for webpy 0.3x. Put this in the same directory as application.py i.e. the web directory for this to work.
More information on how to use web.background:  http://webpy.org/cookbook/background


anandology:
I don't think background.py is a good approach. Tools like celery, rq solves that problems more elegantly. The background.py may go away in this version.

PR #348, opened 2015-11-22
Change datestr() to make output on Windows the same as on Unix

sboden:
A small fix that changes the output of datestr() in utils.py for windows to match the unix/linux output.

### DIFF
--- a/web/utils.py
+++ b/web/utils.py
@@ -897,6 +897,10 @@ def datestr(then, now=None):
         'January  1, 1969'
         >>> datestr(datetime(1970, 6, 1), now=d)
         'June  1, 1970'
+        >>> datestr(datetime(2014, 12, 13), now=d)
+        'December 12, 2014'        
+        >>> datestr(datetime(2014, 12, 6), now=d)
+        'December  6, 2014'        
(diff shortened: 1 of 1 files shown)


jzellman:
My vote is no on this. Adding special characters for specific OS's makes the code much less simple. And my understanding is datestr is intended for displaying dates as strings on a web page, where special characters don't matter. I think a better approach would be to write a wrapper function for specific cases.

anandology:
@jzellman I agree that it is not nice to have platform specific code. But there seems to be an issue here. The `datestr` doesn't work on window because web.py uses not standard directive `%e`, which is not [officially supported][1] by strftime in Python.

Either we switch to displaying the date as "June 03" or consistently display "June 3" on all platforms using standard directives that'll work on all platforms.

[1]: https://docs.python.org/3.6/library/time.html#time.strftime

anandology:
How about replacing it with:

    out = then.strftime('%B %d').replace(" 0", "  ")

That'll make sure we have only one code path and consistent output on all platforms.

jzellman:
@anandology good idea  👍

PR #357, opened 2016-02-18
Added unicode support for binding to unix domain sockets

vabovyan:
If domain socket is provided as unicode, "No socket could be created" is raised


jzellman:
looks good to me, however how would this be handled under python 3?

vabovyan:
@jzellman this will fail miserably, thanks for noticing it, I will take a look and fix it.

iredmail:
Patch not applicable anymore. And no plan to implement this shortly, feel free to reopen this issue with a valid patch. :)

PR #361, opened 2016-03-08
Change dburl2dict to use urlparse and add simple case.

jzellman:
Also added doc test for a mysql db url.

### DIFF
--- a/web/db.py
+++ b/web/db.py
@@ -11,7 +11,7 @@ __all__ = [
   "database", 'DB',
 ]
 
-import time, os, urllib
+import time, os, urllib, urlparse
 try:
     import datetime
 except ImportError:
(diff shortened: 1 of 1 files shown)


web/db.py
```diff
@@ -1148,28 +1148,26 @@ def dburl2dict(url):
...
-    return out
+    dbn, __ = url.split('://', 1)
+    parts = urlparse.urlparse(urllib.unquote(url))
+
+    return {'dbn': dbn,
+            'user': parts.username,
+            'pw': parts.password,
+            'db': parts.path[1:],
+            'host': parts.hostname,
+            'port': parts.port}
```
anandology: You could use `parts.scheme` to find `dbn`. The `url.split` is not required.

Could you please make that change and update the pull request?

jzellman:
Good call. PR has been updated.

anandology:
Thanks @jzellman.

PR #363, opened 2016-03-25
template files without extension

sandeep06011991:
If the template file names do not have an extension like 'file.html' but are named as 'file'
render = web.template.render('templates')
print render.hello('world')
[works only for files named as hello.(some extension)]
 The glob module used  requires unix style wildcards which are not the same as regex .
If it where regex then filename.* would match for filename as well filename.extension
This is not correct for wildcards . 

This could also be a design decision , If so please reply .

### DIFF
--- a/web/template.py
+++ b/web/template.py
@@ -1001,7 +1001,7 @@ class Render:
             raise AttributeError, "No template named " + name            
 
     def _findfile(self, path_prefix): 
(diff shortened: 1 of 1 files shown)


anandology:
Not acceptable. Even `render.hell("world")` match the `templates/hello` file.

PR #364, opened 2016-03-25
Template files without extension 

sandeep06011991:
Adding filename.\* + filename to the fnpath search . This would include files with and without extensions


jzellman:
I'm not entirely sure this is desired. Can you give me an example of when a template file without an extension would be used?

anandology:
Fixed in 123267eff0dd9e46810c817b9e599aa8b6f26c96.

PR #370, opened 2016-06-24
Python 3 full porting

Zashas:
Hi,

This is a PR with a full porting to Python 3. The code has been tested with Python 2.7 and Python 3.5 and both versions passes all unittests. On the other hand, I haven't tested the porting on a real application.

There's one doctest failing with Py2 (minor bug concerning a SyntaxError traceback in templates) that I couldn't fix myself (actually I don't know if it's a bug or a Py2 limitation), otherwise everything should be fine.


anandology:
Quite impressive. Things seems to be working well. 

I figured out how to fix the breaking test in template. 

The reloader seems to be failing. I'll report more as I keep testing it.

anandology:
I've pushed your branch to webpy/webpy and add some more commits to fix the template issue. Also added travis integration for python 3.5 and all the tests are passing now. 

https://travis-ci.org/webpy/webpy/branches

anandology:
There still seems to be issues with str/bytes handling between layers. I was thinking of adding example apps to the repo and write test cases for them and make sure all of them pass. @Zashas what do you think?

Issue #372, opened 2016-06-24
Template sandboxing is not working in Python 3


anandology:
web.py template engine doesn't allow unsafe code in the template by making sure blacklisted nodes are not present in the AST. 

This feature is not working in Python 3 as the `compiler` module that is used by template has been removed in Python 3.

anandology:
The `ast` modules seems to have a parse method, but that seems to be returning a completely different object.

Zashas:
I made a fix for this (cf commit mentioned here above).  The nodes' names changed a bit between the compiler and ast modules, a lot of new ones were added, and I was too lazy to add them all, so I made instead a disallow list. I maybe forgot one or two important node type, so I'd appreciate if someone can double-check this !

anandology:
Having a whitelist of nodes instead of blacklist of nodes was a design decision. I don't think it is a good idea to change that.

PR #375, opened 2016-06-27
Fixes #371 : webapp crashes when unicode text is returned as response

### LINKED ISSUE #371: webapp crashes when unicode text is returned as response
Code to reproduce the bug is available at

https://gist.github.com/anandology/72168ffbf33e87ebcfb419d12fd5e7cd

It crashes with the following error.

	UnicodeEncodeError('latin-1', '\u0c05\u0c06', 0, 2, 'ordinal not in range(256)')
	  File "/Users/anand/github/webpy/webpy/web/wsgiserver/wsgiserver3.py", line 1089, in communicate
	    req.respond()
	  File "/Users/anand/github/webpy/webpy/web/wsgiserver/wsgiserver3.py", line 877, in respond
	    self.server.gateway(self).respond()
	  File "/Users/anand/github/webpy/webpy/web/wsgiserver/wsgiserver3.py", line 1991, in respond
	    chunk = chunk.encode('ISO-8859-1')
	UnicodeEncodeError: 'latin-1' codec can't encode characters in position 0-1: ordinal not in range(256)

cc: @Zashas

Zashas:
In Py3, the data of the response sent to the WSGI server is now always in UTF8-encoded bytes.
I had to modify a bit the Browser class for it to handle this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,7 @@ language: python
 python:
   - "2.6"
   - "2.7"
-#  - "3.1"
-#  - "3.2"
+  - "3.5"
 install: pip install -r test/requirements.txt --use-mirrors; if [ $TRAVIS_PYTHON_VERSION == 2* ] ; then pip install -r test/requirements2.txt --use-mirrors; fi
 script: python test/alltests.py
 before_install:
--- a/test/alltests.py
+++ b/test/alltests.py
@@ -1,7 +1,7 @@
 import webtest
 
 def suite():
-    modules = ["doctests", "db", "application", "session"]
+    modules = ["doctests", "db", "application", "session", "wsgi"]
     return webtest.suite(modules)
     
 if __name__ == "__main__":
--- a/test/application.py
+++ b/test/application.py
@@ -47,17 +47,17 @@ class ApplicationTest(webtest.TestCase):
         import foo
         app = foo.app
         
-        self.assertEquals(app.request('/').data, 'a')
+        self.assertEquals(app.request('/').data, b'a')
         
         # test class change
         time.sleep(1)
         write('foo.py', data % dict(classname='a', output='b'))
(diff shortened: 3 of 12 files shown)


Zashas:
Woops, kinda messed my branch. This PR includes another bugfix. Merging shoudn't be an issue.

Zashas:
I found a bug with this PR using Py2, I'll fix it ASAP.

Zashas:
Fixed. Patch should work correctly.

anandology:
Looks like with this patch, the app will fail if bytes are returned from the app. There are genuine cases where bytes need to be returned like sending back an image etc.

Zashas:
Now all the WSGI layers should use bytes for storing and sending the response. I've written two unittests to ensure that it works. A view can now return either bytes, a string or a unicode string (in Py2) and it should flawlessy be encoded into bytes and properly served. 

This PR also includes a fix for issue #372. I've added a whitelist of AST nodes instead of a blacklist. Yet I couldn't find a proper list of all Python's AST nodes so it's possible false-alerts could be raised in some complex templates. I'd appreciate if someone could test this whitelist with his own templates.

anandology:
Thanks @Zashas. That looks good to me. 

I think it would be hard to merge your branch directly as it is showing some commits that are already there. Either you can create a new branch and cherry-pick your commits and send a new PR or I can cherry-pick all your changes and merge. If I do that, you'll not be able to get the latest changes to your branch. 

Please let me know which way do you prefer.

Zashas:
I prefer if you cherry-pick and merge them. I'll make a new fresh fork.

Thanks !

Zashas:
I've added one more bugfix.

anandology:
@Zashas cherry-picked all your the commits.

Zashas:
Thanks !

I'm closing this.

PR #376, opened 2016-07-01
Extended mapping: more types of URL mappings handled, views can be objects

Zashas:
Two main goals :
- More types of URL mappings are now handled : beside tuples, it now handles a dict or a custom class
- Views can now be general objects (i.e. an instantiated class), and not mandatory a class anymore

This little commit enables more flexibility when writing web.py apps. We actualy need this for our INGInious tool (http://inginious.org/), where our views are splitted across different files, but need access to some constant parameters generated when the application is launched.

See the added tests for some examples of these features.

### DIFF
--- a/web/application.py
+++ b/web/application.py
@@ -111,8 +111,14 @@ class application:
         # Clearing up all thread-local state to avoid interefereing with subsequent requests.
         utils.ThreadedDict.clear_all()
 
+
     def init_mapping(self, mapping):
-        self.mapping = list(utils.group(mapping, 2))
+        if isinstance(mapping, tuple):
+            self.mapping = list(utils.group(mapping, 2))
+        elif hasattr(mapping, "__iter__") and hasattr(mapping, "append"):
+            self.mapping = mapping
+        else:
+            raise Exception("Invalid mapping object. Custom objects should implement at least the methods __iter__ and append.")
 
     def add_mapping(self, pattern, classname):
         self.mapping.append((pattern, classname))
@@ -436,6 +442,15 @@ class application:
                 raise web.nomethod(cls)
             tocall = getattr(cls(), meth)
             return tocall(*args)
+
+        def handle_object(obj):
+            meth = web.ctx.method
+            if meth == 'HEAD' and not hasattr(obj, meth):
+                meth = 'GET'
+            if not hasattr(obj, meth):
+                raise web.nomethod(obj)
+            tocall = getattr(obj, meth)
+            return tocall(*args)
             
         def is_class(o): return isinstance(o, (types.ClassType, type))
             
@@ -463,7 +478,7 @@ class application:
         elif hasattr(f, '__call__'):
             return f()
         else:
-            return web.notfound()
+            return handle_object(f)
 
     def _match(self, mapping, value):
         for pat, what in mapping:
--- a/test/application.py
+++ b/test/application.py
@@ -343,5 +343,66 @@ class ApplicationTest(webtest.TestCase):
         thread.join(timeout=1)
         self.assertFalse(thread.isAlive())
 
+    def test_obj_mapping(self):
+        class Urls(object):
+            mapping = []
+            def __init__(self, mapping):
+                self.mapping = list(web.utils.group(mapping, 2))
+
+            def append(self, uri, f):
+                self.mapping.append((uri, f))
+
+            def __iter__(self):
+                for i in self.mapping:
+                    yield i
+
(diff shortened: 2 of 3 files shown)


web/application.py
```diff
@@ -111,8 +111,16 @@ def _cleanup(self):
         # Clearing up all thread-local state to avoid interefereing with subsequent requests.
         utils.ThreadedDict.clear_all()
 
+
     def init_mapping(self, mapping):
-        self.mapping = list(utils.group(mapping, 2))
+        if isinstance(mapping, tuple):
+            self.mapping = list(utils.group(mapping, 2))
+        elif isinstance(mapping, dict):
+            self.mapping = list(mapping.iteritems())
```
anandology: iteritems doesn't work in python 3.

web/application.py
```diff
@@ -111,8 +111,16 @@ def _cleanup(self):
...
+        elif hasattr(mapping, "__iter__") and hasattr(mapping, "append"):
+            self.mapping = mapping
```
anandology: What is this for?

anandology:
I would love to merge all the INGInious customizations into web.py, but I would like to not have any new features in the py3 release. What do you think?

web/application.py
Zashas: One INGInious dev found that using a simple tuple for mapping URLs was a bit odd, and probably not the most intuitive data structure.

Here we add two new ways of defining a mapping :
- with a dict (see test_dict_mapping)
- with a custom class, which has to be iterable and support an append(url, view) method (see test_obj_mapping)

The dict option is an easy and intuitive data structure and could become the default way of doing this (if we change the docs). The custom class option is needed by INGInious, where each user has a different mapping, built dynamically according to his role/privileges.

Zashas:
I've made this PR just for Py2, assuming that we would fix the iteritems issue when merging the Py3 branch in master.

Indeed the Py3 release is an issue that needs some discussion. I see two possibilities : 
- We either do a pure Py3 release without adding new features. It's not a bad idea, but it has to be done quickly. Otherwise, it penalizes new developing. Here, for instance, I can't make a Py2/Py3 compatible PR outside the Py3 branch because I need the iteritems helper from web/py3helpers.py. This option will need some more testing from users.
- We do a broader release. Since 0.38 is fixed, we can merge Py3 into master now, launch a new roadmap and release Py3 in it later. This facilitates new developing in the short-term.

So it depends how you want to test the Py3 port. If we assume that the py3 port is OK and doesn't need further testing, we can go with option 1, and just say that 0.38 has become Py3-compatible.

By the way, it's the only custom feature INGInious had to made to web.py :)

anandology:
Dictionary is not the right data structure for storing URLs as the order is important. For example, consider the following:

    urls = {
        "/": "index",
        "/.*/edit": "edit_page",
        "/.*": "view_page",
    }
Will /foo/edit will go to `view_page` or `edit_page`? Since we are using a dictionary the behaviour is unpredictable. Where as with the current approach, the behaviour is clearly defined.

    urls = (
        "/", "index",
        "/.*/edit", "edit_page",
        "/.*", "view_page",
    )

I agree that a tuple is not the right data structure for this, probably a list of tuples is. But that is lot of typing and thats why @aaronsw choose this. 

Given the limitation of dict, I'm not sure if we should support it at all. I'm rather thinking that supporting list of tuples instead of dict might be a good idea. With that, people can pass urls.items() to the application.

Zashas:
You're right, the priority issue didn't cross my mind. I've remove the possibility of using dicts from the commit.

anandology:
I'm not very happy with this change, as it might break some existing code. As of now, the following code is valid:

    urls = [
       "/", "index"
    ]
    app = web.applicaiton(urls, globals()

With your changes, it'll start breaking. 

This feature can be very easily supported in the application code by extending the `web.applicaiton` as follows:

    class Application(web.application):
        def init_mapping(self, mapping):
            self.mapping = MyCustomMapping(mapping)

I'm trying to find an example of using objects instead of classes in INGInious. Will update about that feature shortly.

anandology:
I've looked at the INGInious code and how and why the objects are used instead of classes. It looks like that can be solved much more elegantly in other ways. I'll create an issue in  INGInious repo to discuss about it.

anandology:
@Zashas I don't think these changes are required. I've proposed alternative ways to solve the same without making changes to web.py. Please see the conversation in https://github.com/UCL-INGI/INGInious/issues/117

I'm closing this PR without merging. Thanks a lot for your work.

PR #377, opened 2016-07-14
Improving Templetor security feature (thanks egbutter)

Zashas:
This is a merge of egbutter's Templetor modifications with the Py3 port (see https://github.com/webpy/webpy/pull/284).

It improves the security check (attributes are now checked, they weren't before). I added a testsuite instead of egbutter's doctests because of the Py2/Py3 compatibility (doctest is really painful for porting ...)


anandology:
I agree. Even I'm realizing that doctests are very painful. Thinking of moving all the tests to py.test  to make life simpler. 

There is one test using dictionary comprehensions is failing for Python 2.6. Could you please remove/comment-out that, for now?

Zashas:
Done.

PR #387, opened 2016-09-23
Fix: command line should has higher priority than environment variable

chengweiv5:
Environment variable has the highest priority may bring some surprises to user
since user may not configured such PORT environment variable but command line
argument, but he found web.py listen at a unknown port rather than the port
specified by the user.

I think this is why almost of all programs follow this trandition in Linux
world.

Signed-off-by: Chengwei Yang yangchengwei@qiyi.com

### DIFF
--- a/web/wsgi.py
+++ b/web/wsgi.py
@@ -52,9 +52,9 @@ def runwsgi(func):
             return runscgi(func)
     
     
-    server_addr = validip(listget(sys.argv, 1, ''))
-    if 'PORT' in os.environ: # e.g. Heroku
+    if 'PORT' in os.environ: # use PORT if available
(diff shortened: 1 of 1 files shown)


web/wsgi.py
```diff
@@ -52,9 +52,9 @@ def runwsgi(func):
             return runscgi(func)
     
     
-    server_addr = validip(listget(sys.argv, 1, ''))
-    if 'PORT' in os.environ: # e.g. Heroku
```
jzellman: Why was comment changed. How does this change affect existing users? Thinking Heroku since it is mentioned there?

iredmail:
No plan for this change.

PR #389, opened 2016-10-08
Support for ssl_certificate_chain

jlguardi:
I've experienced issues with ssl certificates with intermediates. I've exported the necessary params.

### DIFF
--- a/web/httpserver.py
+++ b/web/httpserver.py
@@ -194,7 +194,7 @@ def WSGIServer(server_address, wsgi_app):
     
     server = wsgiserver.CherryPyWSGIServer(server_address, wsgi_app, server_name="localhost")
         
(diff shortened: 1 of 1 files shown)


jzellman:
looks good to me, however I don't fully understand the implications to existing users. Could this break them? Or is there a chance where the chain is wrong?

jlguardi:
@jzellman No, it doesn't break anything. It allows the use of certificates with intermediates but it doesn't affect to other certificates.

jzellman:
Thanks @jlguardi  @anandology :ship:

jzellman:
Same as #319

anandology:
CherryPy Webserver has been replaced with Cheroot. Please see http://webpy.org/cookbook/ssl for new instructions.

PR #392, opened 2016-12-29
Fix a logical bug.

Hedayet:
"not" has higher precedence over "and" operator. So, expression "not x and y" != "not x or not y".

The target is to check if any of the addr and port are invalid, but the current expression is true when addr is invalid but port is valid.


anandology:
Good catch! Thanks for the fix.

PR #400, opened 2017-02-15
Add the encoding='utf-8' to ensure that it won't cause Chinese disorderly code during the template files are read or wrote

liuming1024:
Today I found a problem when use web.py to develop a web app. The Exception is like this:
```
  File "E:\软件安装\Python\lib\site-packages\web.py-0.40.dev0-py3.6.egg\web\application.py", line 257, in process
    return self.handle()
  File "E:\软件安装\Python\lib\site-packages\web.py-0.40.dev0-py3.6.egg\web\application.py", line 248, in handle
    return self._delegate(fn, self.fvars, args)
  File "E:\软件安装\Python\lib\site-packages\web.py-0.40.dev0-py3.6.egg\web\application.py", line 488, in _delegate
    return handle_class(cls)
  File "E:\软件安装\Python\lib\site-packages\web.py-0.40.dev0-py3.6.egg\web\application.py", line 466, in handle_class
    return tocall(*args)
  File "F:\liuming\PyWorkspace\WeChat\QuickStart.py", line 56, in GET
    return render.form(name)
  File "E:\软件安装\Python\lib\site-packages\web.py-0.40.dev0-py3.6.egg\web\template.py", line 1034, in __getattr__
    t = self._template(name)
  File "E:\软件安装\Python\lib\site-packages\web.py-0.40.dev0-py3.6.egg\web\template.py", line 1031, in _template
    return self._load_template(name)
  File "E:\软件安装\Python\lib\site-packages\web.py-0.40.dev0-py3.6.egg\web\template.py", line 1016, in _load_template
    return Template(open(path).read(), filename=path, **self._keywords)
UnicodeDecodeError: 'gbk' codec can't decode byte 0xbd in position 139: illegal multibyte sequence
```
And this exception happens because the following code:
```
  File "E:\软件安装\Python\lib\site-packages\web.py-0.40.dev0-py3.6.egg\web\template.py", line 1016, in _load_template
    return Template(open(path).read(), filename=path, **self._keywords)
```

So, I change `return Template(open(path).read(), filename=path, **self._keywords)` to `return Template(open(path,encoding='utf-8').read(), filename=path, **self._keywords)`, then the problem was sovled. 
And the Chinese disorderly code problem was also solved:
Before>>
Chinese disorderly code
After>>

### DIFF
--- a/web/template.py
+++ b/web/template.py
@@ -925,7 +925,7 @@ class Template(BaseTemplate):
 
         def get_source_line(filename, lineno):
             try:
-                lines = open(filename).read().splitlines()
+                lines = open(filename, encoding='utf-8').read().splitlines()
                 return lines[lineno]
             except:
                 return None
@@ -1013,7 +1013,7 @@ class Render:
         if kind == 'dir':
             return Render(path, cache=self._cache is not None, base=self._base, **self._keywords)
         elif kind == 'file':
-            return Template(open(path).read(), filename=path, **self._keywords)
+            return Template(open(path, encoding='utf-8').read(), filename=path, **self._keywords)
         else:
             raise AttributeError("No template named " + name)
 
@@ -1075,7 +1075,7 @@ except ImportError:
 def frender(path, **keywords):
     """Creates a template from the given file path.
     """
-    return Template(open(path).read(), filename=path, **keywords)
+    return Template(open(path, encoding='utf-8').read(), filename=path, **keywords)
     
 def compile_templates(root):
     """Compiles templates to python code."""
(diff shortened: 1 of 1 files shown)


anandology:
Looks good to me. I think you accidentally added a VCS.xml file. Could you
please remove that?

i2it:
@anandology Sure, now I have removed VCS.xml file.

jzellman:
@anandology thoughts on .gitignore change? It is separate commit... otherwise LGTM

anandology:
Doesn't work with python 2.

```
$ python2.7
>>> open('/tmp/a.txt', encoding='utf-8')
  File "<stdin>", line 1, in <module>
TypeError: 'encoding' is an invalid keyword argument for this function
```

To make it work with Python 2, we need to import the `open` function from `io` module.

```
>>> from io import open
>>> open('/tmp/a.txt', encoding='utf-8')
<_io.TextIOWrapper name='/tmp/a.txt' encoding='utf-8'>
```

@i2it could you please make that change and remove the change the .gitignore?

anandology:
Merged in ffb0b6a44ac4ed51642e693c1bc7a215b67223eb.

PR #407, opened 2017-02-19
Issue380

eashan:
Added Support for mysql-connector Driver

### DIFF
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -3,4 +3,4 @@ MySQL-python
 PyGreSQL
 DBUtils
 PyMySQL
-MySQLConnector
+mysql-connector
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -3,4 +3,3 @@ MySQL-python
 PyGreSQL
 DBUtils
 PyMySQL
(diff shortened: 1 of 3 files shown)


test/requirements.txt
```diff
@@ -2,4 +2,5 @@ psycopg2
 MySQL-python
 PyGreSQL
 DBUtils
-PyMySQL
\ No newline at end of file
+PyMySQL
+MySQLConnector
```
anandology: MySQLConnector doesn't seem to be a valid python package.

anandology: Please have a look at the following discussion on stackoverflow to see how to install it using pip.

http://stackoverflow.com/questions/22100757/can-not-get-mysql-connector-python-to-install-in-virtualenv

anandology:
Please look at the [travis log][1], which is linked in the "show all checks" link above. It has the error message that caused the pip install to fail.

[1]: https://s3.amazonaws.com/archive.travis-ci.org/jobs/203114113/log.txt

anandology:
Thanks @eashan. Merged.

PR #413, opened 2017-03-11
Updates .gitignore

vharsh:
The configuration now ignores Python eggs, build and dist directories.


anandology:
Thanks @varunpurohit76.

PR #419, opened 2017-05-12
import SimpleCookie for compatibility with both python 2 and 3

jamieforth:
Simple fix to import SimpleCookie rather than the previous Cookie.SimpleCookie().

Also switched from cookie.iteritems() to cookie.items.


anandology:
Thanks @jamieforth. Could you please add a test case?

cclauss:
Is this fixed?

PR #427, opened 2017-07-08
Templates load error in Python3

Garfield0428:
Templates load error in Python3
_2_9 20xx wf wg x s
p _ 8 v o84 1mr5e4 ouxh

I fix it by adding  encoding='utf-8'


anandology:
Already fixed in #400.

PR #98, opened 2011-09-14
HTML5 Forms

faruken:
Some HTML5 forms such as email, url, number, range, color, search, telephone and datalist.

PR #440, opened 2017-09-17
allow HTML5 form input elements

jimgregory:
HTML5 now allows for 13 additional <input /> elements from the original HTML4 spec, like 'email', 'tel', 'date', etc.  All modern browsers support them, and they give a better user experience with intuitive widgets and basic client-side error checking.

Each new input type is defined by declaring a type attribute on an input element, e.g. `<input name="foo" type="date"/>`.  Rather than creating specific classes for all 13 of them, which would be difficult to remember and greatly expand the code base, I propose modifying the generic Input element so it allows a type element to be passed as an input to the class; i.e. `web.form.Input(name="foo", type="date")`.  This will prevent having to update form.py in the future if/when other new input types are added.

I did not include any check to see if the type is a valid HTML5 form type.  I can include a list of valid types to check against, but this would again become invalid if/when the set of valid HTML5 input types is extended.

### DIFF
--- a/web/form.py
+++ b/web/form.py
@@ -122,6 +122,11 @@ class Form(object):
     d = property(_get_d)
 
 class Input(object):
+    """Generic input. Type attribute must be specified when called directly
+
+        >>> Input(name='foo', value='bar', type='number').render()
+        u'<input id="foo" name="foo" value="bar" type="number"/>'
+    """
     def __init__(self, name, *validators, **attrs):
         self.name = name
         self.validators = validators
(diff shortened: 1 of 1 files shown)


web/form.py
```diff
@@ -143,7 +149,10 @@ def is_hidden(self):
         return False
         
     def get_type(self):
-        raise NotImplementedError()
+        if self.type != None:
+            return self.type
+        else:
+            raise AttributeError("missing attribute 'type'")
```
blueyed: Why not just `return self.type`?  (would also raise AttributeError).
Anyway: `!= None` should be `is not None`.

blueyed: But in general I think
```python
try:
    return self.type
except AttributeError:
    return None
```
would be best.

iredmail:
This PR should replace https://github.com/webpy/webpy/pull/98

I will merge it after upcoming stable release.

PR #441, opened 2017-10-07
Remove the warnings during python3 setup.py install process

pierrecontri:
Dear,

During the python3 setup.py install, I have 5 warnings due to the "except socket.error, e:" and a raise array[0], array[1], array[2].

file wsgiserver.py, line 1019, 1040, 1035, 1095, 2285

I transform this line on a tuple for except, and no warnings.
But the real reflexion is "what the python3 setup.py install command use the wsgiserver2.py ?"
I have no time to investigate, so sorry.

Best Regards, and many thanks for the Great Tool !
Peter

### DIFF
--- a/web/wsgiserver/wsgiserver2.py
+++ b/web/wsgiserver/wsgiserver2.py
@@ -1016,7 +1016,7 @@ class CP_fileobject(socket._fileobject):
             try:
                 bytes_sent = self.send(data)
                 data = data[bytes_sent:]
-            except socket.error, e:
+            except (socket.error, e):
                 if e.args[0] not in socket_errors_nonblocking:
                     raise
 
@@ -1037,7 +1037,7 @@ class CP_fileobject(socket._fileobject):
                 data = self._sock.recv(size)
                 self.bytes_read += len(data)
                 return data
-            except socket.error, e:
+            except (socket.error, e):
                 if (e.args[0] not in socket_errors_nonblocking
                         and e.args[0] not in socket_error_eintr):
                     raise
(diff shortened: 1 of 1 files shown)


web/wsgiserver/wsgiserver2.py
```diff
@@ -1016,7 +1016,7 @@ def sendall(self, data):
             try:
                 bytes_sent = self.send(data)
                 data = data[bytes_sent:]
-            except socket.error, e:
+            except (socket.error, e):
```
songww: In py3, It must be `except socket.error as e`(works in py2 too), not `except (socket.error, e)`. 
`except (socket.error, e)` means except two Exception: socket.error and e, but e is an instance of `socket.error`.

pierrecontri:
Dear,

Ok, so sorry for this proposal.

But I have a warning on my python3 setup.py install.

I didn't take time to analyse the code more details.

for me the initial code "except socket.error, e:" seems that there is 2 
exceptions one defined with a specific type and the second (else except) 
get the rest of exceptions into the e instance.

So excuse me for the pull request.

I'll have a look on this error (and not warning like I wrote previously) 
on my installation and try to correct with "except socket.error as e:".

But the real question is why on the python3 installation the 
wsgiserver2.py is called and not the wsgiserver3.py ?

No errors actually with "except socket.error as e:"

Thanks a lot for your reply and congratulation for your great code.

Best Regards,
Peter


Le 10/10/2017 à 12:28, songww a écrit :

anandology:
I don't think this happens any more with the master branch.

PR #442, opened 2017-10-10
see https://github.com/webpy/webpy/pull/400

liuming1024:


anandology:
Thanks @liuming1024. Both #400 and #442 are merged now.

PR #451, opened 2018-04-25
Feature route decorator

akash0x53:
Added Flask like route decorator

Example:

```
import web
app = web.application()
@app.route("/hello")
def hello():
    return "Hello World"

app.run()
```

Any thoughts, suggestions to make Web.Py modern.

### DIFF
--- a/tests/test_routing.py
+++ b/tests/test_routing.py
@@ -1,0 +1,47 @@
+import unittest
+import time
+import threading
+import sys
+import web
+import urllib
+
+try:
+    from urllib.parse import urlencode
+except ImportError:
+    from urllib import urlencode
+
+from web.py3helpers import PY2
+
+
+class RouteTest(unittest.TestCase):
+    def testHandlerMethod(self):
+        app = web.application()
+        
+        @app.route('/method')
+        def handler_method():
+            return 'Hello Method'
+            
+        response = app.request('/method')
+        self.assertEquals(response.data, 'Hello Method')
+    
+    def testHandlerClass(self):
+        app = web.application()
+        
+        @app.route('/class')
+        class HandlerClass:
+            def GET(self):
+                return 'Hello Class'
+            
+        response = app.request('/class')
+        self.assertEquals(response.data, 'Hello Class')
+    
+    def testParameters(self):
+        app = web.application()
+        
+        @app.route('/hello/(.*)')
+        class HandlerClass():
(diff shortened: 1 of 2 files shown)


iredmail:
This feature is cool, I will merge it after upcoming stable release. :)

tests/test_routing.py
```diff
@@ -0,0 +1,47 @@
+import unittest
+import time
+import threading
+import sys
+import web
+import urllib
```
cclauss: OPTIONAL: run [isort](http://isort.readthedocs.io) on the imports.

iredmail: Seems `isort` is not Black compatible.

```
$ black --diff db.py
--- db.py	2019-09-22 02:42:22.727265 +0000
+++ db.py	2019-09-22 02:42:36.098191 +0000
@@ -8,12 +8,11 @@
 import os
 import re
 import time
 
 from .py3helpers import PY2, iteritems, numeric_types, string_types, text_type
-from .utils import (iterbetter, iters, safestr, safeunicode, storage,
-                    threadeddict)
+from .utils import iterbetter, iters, safestr, safeunicode, storage, threadeddict
```

`isort` generates the `-` lines, but Black generates `+` lines.

cclauss: __isort --line_length 88__

iredmail: Cool, didn't realize the `--line_length` argument.
Could you like to help review and merge this? This makes web.py familiar with Flask.

anandology:
@iredmail and @cclauss I think this is a quite a big change to the API and requires more discussion. I suggest not rush when making major changes to the API. 

In fact, @aaronsw and I've used this pattern in Infogami and didn't feel that it would should made the default API in web.py. One of the drawbacks of this approach is that it becomes hard to see what are URLs are served by the application. Also, web.py uses regular expressions for pattern matching where are Flask uses rule based matching like `/posts/<int:post_id>`. Just adding a decorator without thinking about all other implications would make it very inconsistent API and there will be more than one way of doing things and it would be hard for users to decide which one to use when. 

I prefer not to have any major API changes for 0.40.0 release.

iredmail:
> One of the drawbacks of this approach is that it becomes hard to see what are URLs are served by the application.

Agree to this point. Each time I tried Flask I was struggling with the route decorator. 

Let’s postpone this when there’s more demand.

PR #452, opened 2018-04-29
fix DeprecationWarning: generator 'IterBetter.__iter__' raised StopIteration for Python 3.5+

gaopinghuang0:
Issue: DeprecationWarning: generator 'IterBetter.__iter__' raised StopIteration

My change is based on the example from [PEP 479: Change StopIteration handling inside generators](https://www.python.org/dev/peps/pep-0479/#examples-of-breakage)
OLD:
```python
if context is None:
    while True:
        yield next(line_pair_iterator)
```
Becomes:
```python
if context is None:
    for line in line_pair_iterator:
        yield line
    return
```

### DIFF
--- a/web/utils.py
+++ b/web/utils.py
@@ -664,10 +664,12 @@ class IterBetter:
     def __iter__(self): 
         if hasattr(self, "_head"):
             yield self._head
-
-        while 1:    
-            yield next(self.i)
+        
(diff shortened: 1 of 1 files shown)


anandology:
Fixed by #459.

anandology:
@gaopinghuang0 thanks for the PR anyway.

PR #458, opened 2018-06-27
Use public domain dedication Unlicense

TrashMacNugget:
In some jurisdictions there are complications putting software into the public domain as well as using it. The Unlicense dedicates the software to the public domain, but also explicitly grants you the rights in jurisdictions where putting things in the public domain isn't applicable.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,1 +1,24 @@
-web.py is in the public domain; it can be used for whatever purpose with absolutely no restrictions.
+This is free and unencumbered software released into the public domain.
+
+Anyone is free to copy, modify, publish, use, compile, sell, or
+distribute this software, either in source code form or as a compiled
+binary, for any purpose, commercial or non-commercial, and by any
+means.
+
+In jurisdictions that recognize copyright laws, the author or authors
+of this software dedicate any and all copyright interest in the
+software to the public domain. We make this dedication for the benefit
+of the public at large and to the detriment of our heirs and
+successors. We intend this dedication to be an overt act of
+relinquishment in perpetuity of all present and future rights to this
+software under copyright law.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+OTHER DEALINGS IN THE SOFTWARE.
+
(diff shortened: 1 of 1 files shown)


anandology:
I agree with @sbp that Unlicense is not the way to go. 

CC0 looks likes a good option. Are they any things to be worried about if we use that license? Do we need to get some kind of consent from the past/future contributors? @sbp any ideas?

sbp:
@anandology If a public domain dedication could be changed to CC0 without consequence then presumably there would be no reason to do so in the first place. You could just keep the existing dedication.

Any code using a public domain dedication could, perhaps, be relicensed using the following mechanism. The code could first be relicensed to CC0 in a jurisdiction that allows such an action (i.e. in any jurisdiction that accepts a public domain dedication), creating a fork. This fork can then be used globally. The original project could legally clone the fork back, possibly in the same jurisdiction if required, thereby importing the new license. This would be the legal equivalent of [circular reporting](https://en.wikipedia.org/wiki/Circular_reporting) or [citogenesis](https://xkcd.com/978/). Unfortunately as I am not a lawyer I do not know in which jurisdictions such actions would be valid.

Since web.py has no [Contributor License Agreement](https://en.wikipedia.org/wiki/Contributor_License_Agreement) (CLA) in place it could also be argued that only the original code is public domain, and that no subsequent contributions have a license. Practically speaking, I think that if somebody issues a pull request to a project with a specific license then they understand that their contributions will be incorporated under that license.

You may find, however, that some large organisations are nervous about such point of view and therefore ask for more explicit confirmation from contributors. The Apache Software Foundation, for example, specifically include a CLA in the Apache License 2.0 (in Section 5) for this reason. But in addition they also make you sign a [separate CLA](https://www.apache.org/licenses/icla.pdf) when you join as a committer. The Linux Kernel maintainers use a slightly different mechanism called the [Developer Certificate of Origin](https://developercertificate.org/) that all contributors have to assert. This contains a clause about publication of metadata in Section (d), which may be made more relevant in the age of GDPR.

Some big software projects have their own custom CLAs. When contributing to jQuery for example you do so under the [JS Foundation CLA](https://js.foundation/CLA), and when contributing to CPython you do so under the [PSF's CLA](https://www.python.org/psf/contrib/). When contributing to Go you must do so under [Google's CLA](https://cla.developers.google.com/about/google-individual). Interestingly Google's CLA includes a patent grant, somewhat like the Apache License 2.0, but separate from and hence in addition to any individual license.

To appease the pedants of this world, my citogenesis jurisdiction relicensing suggestion plus the inclusion of the DCO 1.1 may be sufficient. But I of course recommend checking with a lawyer.

TrashMacNugget:
My only concern is that currently the license just says "this is public domain" without much more, which is not legally thorough. CC0 and Unlicense give you the same rights, so I think it is fine to use either.

I personally prefer the Unlicense, as it is short and I think stays with the spirit of the original public domain dedication, but if there is any concern, CC0 can be used instead. If you have a preference for a certain license, the CC0 and Unlicense give you the right to relicense and use the source under really whatever license you want, so I don't think it matters what license is chosen.

TrashMacNugget:
I don't think getting permission to relicense from all contributors will be needed because even if they are in a jurisdiction that doesn't allow you to place work into the public domain, the license _sort of_ has a fallback.

> web.py is in the public domain; **it can be used for whatever purpose with absolutely no restrictions.**

I think the second clause give you these rights regardless of whether the first clause is true. But I'm not a lawyer.

PR #459, opened 2018-07-05
fixed on Python 3.7 a RuntimeError caused a StopIteration while iterating through a different iterator in a generator

costypetrisor:


anandology:
Thanks @costypetrisor. Could you please add a test case that would fail without this fix?

costypetrisor:
I don't feel we need to add a test for this one. Existing tests in this codebase already failed because of this problem.

Most generators must end at some point (unless you intentionally make an infinite one like `itertools.cycle`) and this is the new norm on Python 3.7.

But it's a good thing you insisted because yesterday I was in a rush and didn't run the tests for this codebase. Now I've did run them and found the same problem in 2 more places, so I've updated my MR accordingly  :) .

costypetrisor:
@rafalstapinski I liked your patch, so hope you don't mind that I cherry-picked your exact commit on top of my branch, just so that the maintainer will have less to do in order to integrate

anandology:
This PR is merged now.

Thanks @costypetrisor  and @rafalstapinski .

PR #460, opened 2018-07-09
Dual-license under public domain dedications CC0+Unlicense

TrashMacNugget:
In some jurisdictions there are complications putting software into the public domain as well as using it.

This fixes the concerns raised in #458 with just using the Unlicense by dual-licensing both it and the CC0, letting downstream choose which one to use. It also deals with my concerns with CC0 (patents, length), because if downstream is comfortable with the Unlicense they are allowed to choose it.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,1 +1,154 @@
-web.py is in the public domain; it can be used for whatever purpose with absolutely no restrictions.
+web.py is in the public domain; it can be used for whatever purpose with
+absolutely no restrictions. You may use it under the terms of the Unlicense [1],
+or the Creative Commons Zero (CC0 1.0) [2] license, at your option.
+
+[1] Unlicense
+=============
+
+This is free and unencumbered software released into the public domain.
+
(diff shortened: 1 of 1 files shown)


sbp:
I'm still in the process of getting legal analysis of #458 (and, hence, now #460 too). The Unlicense is generally not favoured, and to prevent license proliferation dual-licensing CC0 with the Apache License 2.0 would probably make more sense for those concerned about patents. But there are many other considerations too, some of which were documented in #458, and accepting this pull request cannot happen without having considered and accounted for all of the consequences.

cclauss:
👍 on dual-licensing CC0 with the Apache License 2.0

iredmail:
Ping @anandology

anandology:
I would prefer to focus the efforts on getting 0.40 out and look into this
after that.

iredmail:
Any update on the license change?
Do we release new version under public domain?

@sbp @anandology @vclauss

cclauss:
I added the 1.0 milestone in alignment with https://github.com/webpy/webpy/pull/460#issuecomment-531466948

PR #463, opened 2018-07-21
Fix StopIteration error in utils for Python 3.7.0

rafalstapinski:
In Python 3.7.0 implicitly breaking on StopIteration is no longer valid and transforms into a RuntimeError. This change now uses a simpler generator without the need to yield and rely on a StopIteration exception.


anandology:
Thanks @rafalstapinski already merged this as part of #459.

PR #473, opened 2018-09-26
wsgi: allow to run via socket

gmelikov:
Main part of code and modules already support it,
just check input args appropriately.

Signed-off-by: George Melikov <mail@gmelikov.ru>

### DIFF
--- a/web/httpserver.py
+++ b/web/httpserver.py
@@ -168,10 +168,13 @@ def runsimple(func, server_address=("0.0.0.0", 8080)):
     
     server = WSGIServer(server_address, func)
 
-    if server.ssl_adapter:
-        print("https://%s:%d/" % server_address)
+    if '/' in server_address[0]:
+        print ("%s" % server_address)
     else:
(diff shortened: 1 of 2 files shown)


gmelikov:
This change allows us to create docker plugins on webpy.

web/httpserver.py
```diff
@@ -168,10 +168,13 @@ def runsimple(func, server_address=("0.0.0.0", 8080)):
     
     server = WSGIServer(server_address, func)
 
-    if server.ssl_adapter:
-        print("https://%s:%d/" % server_address)
+    if '/' in server_address[0]:
```
anandology: Wouldn't it be better to use `server_address.startswith('/')`?

gmelikov: @anandology i've got this variant from https://github.com/webpy/webpy/blob/master/web/net.py#L156 , but i won't mind to change it, if you want.

Thank you for fast review!

anandology:
Thanks @gmelikov.

PR #484, opened 2018-10-30
Remove trailing whitespace

blueyed:
Used:

    sed -i 's/ \+$//' **/*.py

### DIFF
--- a/experimental/background.py
+++ b/experimental/background.py
@@ -19,7 +19,7 @@ def background(func):
                 if k not in ['status', 'headers', 'output']:
                     try: del myctx[k]
                     except KeyError: pass
-        
+
         t = threading.Thread(target=newfunc)
         background.threaddb[id(t)] = t
         t.start()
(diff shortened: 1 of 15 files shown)


blueyed:
Is this something that is wanted?
I can re-do it then again to fix conflicts, but feel free to do it yourself, of course.

cclauss:
Please fix conflicts.

blueyed:
Would just redo it.
Would prefer to do it after https://github.com/webpy/webpy/pull/485 to ensure there are no unexpected things happening.

blueyed:
Rebased/redone.

cclauss:
There seems to be a better way to achieve this that is built into every installation of CPython...
* https://github.com/python/cpython/blob/master/Tools/scripts/reindent.py

blueyed:
@cclauss 
It is not about reformatting, but only about removing trailing whitespace in one step (i.e. a minimal change).

cclauss:
Next step would be to pass the codebase thru https://GitHub.com/python/black

blueyed:
@cclauss 
Won't do it myself - not using webpy anymore.

iredmail:
@cclauss reformatting is not urgent. Maybe we should release a new stable version first, then drop py2 support since it will be EOF on Jan 1, 2020, get rid of the burden and move on.

PR #487, opened 2018-11-20
Unicode compatability for Python 2-3

DrFaustie:


DrFaustie:
Some issues with compatibility still. Though strangely, my local tests are passing

web/webapi.py
```diff
@@ -409,6 +409,13 @@ def decode_cookie(value):
     >>> decode_cookie('foo \xE9 bar')
     u'foo \xe9 bar'
     """
+    # Py3
+    try:
+        unicode("", "utf-8")
```
cclauss: There is no need to call the function this line can just be: __unicode__

PR #494, opened 2019-01-22
Update LICENSE.txt to CC0

MarkAtwood:
https://creativecommons.org/share-your-work/public-domain/cc0/

The Problem

Dedicating works to the public domain is difficult if not impossible for those wanting to contribute their works for public use before applicable copyright or database protection terms expire. Few if any jurisdictions have a process for doing so easily and reliably. Laws vary from jurisdiction to jurisdiction as to what rights are automatically granted and how and when they expire or may be voluntarily relinquished. More challenging yet, many legal systems effectively prohibit any attempt by these owners to surrender rights automatically conferred by law, particularly moral rights, even when the author wishing to do so is well informed and resolute about doing so and contributing their work to the public domain.

A Solution

CC0 helps solve this problem by giving creators a way to waive all their copyright and related rights in their works to the fullest extent allowed by law. CC0 is a universal instrument that is not adapted to the laws of any particular legal jurisdiction, similar to many open source software licenses. And while no tool, not even CC0, can guarantee a complete relinquishment of all copyright and database rights in every jurisdiction, we believe it provides the best and most complete alternative for contributing a work to the public domain given the many complex and diverse copyright and database systems around the world.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,1 +1,124 @@
 web.py is in the public domain; it can be used for whatever purpose with absolutely no restrictions.
+
+Creative Commons Legal Code
+
+CC0 1.0 Universal
+
+    CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES NOT PROVIDE
+    LEGAL SERVICES. DISTRIBUTION OF THIS DOCUMENT DOES NOT CREATE AN
+    ATTORNEY-CLIENT RELATIONSHIP. CREATIVE COMMONS PROVIDES THIS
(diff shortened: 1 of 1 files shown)


sbp:
In #458 somebody suggested to use the Unlicense, and in #460 the same person suggested to dual license under CC0 and the Unlicense. At the time I identified dual-licensing CC0 and the Apache License 2.0 to be the best approach to cover copyright and patent issues, but I sought proper legal analysis (and am still in the process of doing that) to confirm that idea. In the process I discovered that there may be a better option available to us in the near future, which I'm still currently discussing. Stay tuned!

PR #497, opened 2019-03-17
Implemented Searchbox form-input type with datalist (https://develope…

gjdv:
…r.mozilla.org/en-US/docs/Web/HTML/Element/input/search)


iredmail:
Hi @gjdv Is this still necessary?

gjdv:
I need the Searchbox in my usage and that is why I added it. The code runs fine, not sure why Travis fails (on other things than my addition) for any other python version then 3.5.
I see that the main branch developed quite a bit further since my addition. When I find the time I'll try and pull them into my branch

iredmail:
Please regenerate then patch based on latest git version of web.py, I’m more than happy to merge it and ship it in upcoming stable release.

iredmail:
Here’s another Datalist implementation, it would be better if you can check whether each anything better you can use. 
https://github.com/webpy/webpy/pull/98/files

gjdv:
The implementation from [#98](https://github.com/webpy/webpy/pull/98) also works for me. These are equivalent:

'mySB = form.Searchbox(name='foo', value='bar', args=["bar1", "bar2", "bar3"], placeholder="Type or choose").render()'

'theirSB = form.Datalist(name='foo_list', args=["bar1", "bar2", "bar3"]).render()
theirSB += form.Search(name='foo', value='bar', list='foo_list', placeholder="Type or choose").render()'

I'd only suggest to close the '<option>' tags, see [my comment](https://github.com/webpy/webpy/pull/98/files#r325384786)

cclauss:
Conflicts: This PR needs to be re-based.

iredmail:
@gjdv `Searchbox` is now `Search` in recent commit.

gjdv:
Thanks

Issue #519, opened 2019-09-14
Make sure all web/*.py files match PEP8 coding style


iredmail:
Make sure all web/*.py files match PEP8 coding style

- [x] `__init__.py`
- [x] application.py
- [x] browser.py
- [x] db.py
- [x] debugerror.py
- [x] form.py
- [x] http.py
- [x] httpserver.py
- [x] net.py
- [x] py3helpers.py
- [x] session.py
- [x] template.py
- [x] test.py
- [x] utils.py
- [x] webapi.py
- [x] webopenid.py
- [x] wsgi.py

cclauss:
Let’s jus run Th Python Software Foundation’s Black on the whole repo.  I will submit a PR in a few hours.

iredmail:
@cclauss Maybe stored the black config file in webpy git repo, so that we have same setting.

cclauss:
Black does not take configuration.  This is done on _purpose_.  Everyone gets the same defaults —You can have any color you want as long as it is black.

iredmail:
Quote from its [document](https://black.readthedocs.io/en/stable/pyproject_toml.html):

> Black is able to read project-specific default values for its command line options from a pyproject.toml file.

I suppose you mean we just use all Black's default settings. I'm ok with this, but better mention this in README file or somewhere else.

cclauss:
We will have a pre-commit file so that flake8 tests and black formatting can be auto-run locally in the commit process.

iredmail:
Cool. I will wait for the changes. :)

iredmail:
BTW, i personally dislike Black style function arguments (either definition or calling) like this:

```
            conn.delete(
                "settings",
                vars={"account": account, "params": _params},
                where="account=$account AND k IN $params",
            )
```

Preferred one is one argument per line like below, easier to read:

```
            conn.delete("settings",
                        vars={"account": account, "params": _params},
                        where="account=$account AND k IN $params")
```

Kind of struggling.

iredmail:
@cclauss do we have other options besides Black?

cclauss:
Please read the readme for Black.  This formatting encourages fewer parameters which makes for safer code.  The idea behind Black is to surrender your personal pet peeves in favor of a single standard.  gofmt has worked wonders for Golang and the PSF is pushing for similar unity.  We have more important things to worry about.

PR #520, opened 2019-09-14
WIP: Reformat the code using psf/black

cclauss:
Work in Progress:  Need to add appropriate documentation before merging to master.

Fixes: #519

### DIFF
--- a/experimental/untwisted.py
+++ b/experimental/untwisted.py
@@ -8,27 +8,28 @@ import simplejson
 
 import web
 
+
 class Request(http.Request):
     def process(self):
         self.content.seek(0, 0)
         env = {
-          'REMOTE_ADDR': self.client.host,
-          'REQUEST_METHOD': self.method,
-          'PATH_INFO': self.path,
-          'CONTENT_LENGTH': web.intget(self.getHeader('content-length'), 0),
-          'wsgi.input': self.content
+            "REMOTE_ADDR": self.client.host,
+            "REQUEST_METHOD": self.method,
+            "PATH_INFO": self.path,
+            "CONTENT_LENGTH": web.intget(self.getHeader("content-length"), 0),
+            "wsgi.input": self.content,
         }
-        if '?' in self.uri:
-            env['QUERY_STRING'] = self.uri.split('?', 1)[1]
+        if "?" in self.uri:
+            env["QUERY_STRING"] = self.uri.split("?", 1)[1]
 
         for k, v in self.received_headers.iteritems():
-            env['HTTP_' + k.upper()] = v
+            env["HTTP_" + k.upper()] = v
 
-        if self.path.startswith('/static/'):
-            f = web.lstrips(self.path, '/static/')
-            assert '/' not in f
-            #@@@ big security hole
-            self.write(open('static/' + f).read())
+        if self.path.startswith("/static/"):
+            f = web.lstrips(self.path, "/static/")
+            assert "/" not in f
+            # @@@ big security hole
+            self.write(open("static/" + f).read())
             return self.finish()
 
         web.webapi._load(env)
@@ -38,8 +39,9 @@ class Request(http.Request):
         for (h, v) in web.ctx.headers:
             self.setHeader(h, v)
         self.write(web.ctx.output)
-        if not web.ctx.get('persist'):
+        if not web.ctx.get("persist"):
             self.finish()
+
 
 class Server(http.HTTPFactory):
     def __init__(self, func):
@@ -49,24 +51,33 @@ class Server(http.HTTPFactory):
         """Generate a channel attached to this site.
         """
         channel = http.HTTPFactory.buildProtocol(self, addr)
+
         class MyRequest(Request):
             actualfunc = staticmethod(self.func)
+
         channel.requestFactory = MyRequest
         channel.site = self
         return channel
+
 
 def runtwisted(func):
     reactor.listenTCP(8086, Server(func))
     reactor.run()
 
+
 def newrun(inp, fvars):
     print("Running on http://0.0.0.0:8086/")
     runtwisted(web.webpyfunc(inp, fvars, False))
 
+
 def iframe(url):
-    return """
+    return (
+        """
     <iframe height="0" width="0" style="display: none" src="%s"/></iframe>
-    """ % url #("http://%s.ajaxpush.lh.theinfo.org:8086%s" % (random.random(), url))
+    """
+        % url
+    )  # ("http://%s.ajaxpush.lh.theinfo.org:8086%s" % (random.random(), url))
+
 
 class Feed:
     def __init__(self):
@@ -82,6 +93,7 @@ class Feed:
         for x in self.sessions:
             x.write(text)
 
+
 class JSFeed(Feed):
     def __init__(self, callback="callback"):
         Feed.__init__(self)
@@ -89,9 +101,12 @@ class JSFeed(Feed):
 
     def publish(self, obj):
         web.debug("publishing")
-        Feed.publish(self,
-          '<script type="text/javascript">window.parent.%s(%s)</script>' % (self.callback, simplejson.dumps(obj) +
-          " " * 2048))
+        Feed.publish(
+            self,
+            '<script type="text/javascript">window.parent.%s(%s)</script>'
+            % (self.callback, simplejson.dumps(obj) + " " * 2048),
+        )
+
 
 if __name__ == "__main__":
     mfeed = JSFeed()
@@ -106,7 +121,8 @@ if __name__ == "__main__":
 
     class view:
         def GET(self):
-            print("""
+            print(
+                """
 <script type="text/javascript">
 function callback(item) {
   document.getElementById('content').innerHTML += "<p>" + item + "</p>";
@@ -123,7 +139,8 @@ function callback(item) {
   <input type="submit" value="send" />
 </form>
 <iframe id="foo" height="0" width="0" style="display: none" src="/js"/></iframe>
-            """)
+            """
+            )
 
     class js:
         def GET(self):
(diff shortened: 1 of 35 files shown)


tests/test_application.py
```diff
@@ -24,242 +24,253 @@ def GET(self):
...
-        response = app.request('/', method='internal')
-        self.assertEqual(response.status, '405 Method Not Allowed')
+            def internal(self):
+                return "secret"
+
+        response = app.request("/", method="internal")
+        self.assertEqual(response.status, "405 Method Not Allowed")
 
     def testRedirect(self):
         urls = (
```
iredmail: This format is really ugly.  :(

cclauss: The idea behind Black is to surrender your personal pet peeves in favor of a single standard.

iredmail: I understand the `surrender` part completely, what i mean is readability, not personal favor.

With `urls` like below is easier to read and understand IMO:

```
urls = (
    '<pattern>', 'class-or-func',
    '<pattern>', 'class-or-func',
    '<pattern>', 'class-or-func',
    '<pattern>', 'class-or-func',
    '<pattern>', 'class-or-func',
)
```

cclauss: Then let’s create a __namedtuple__ and write them that way.  The pattern, class_or_func thing is not obvious from reading the current code.

iredmail:
I'm ok with this PR, you can merge it.
Just not yet accept the ugly parts which reduce the readability.

iredmail:
in `.pre-commit-config.yaml`, is it better to use py2.7 as default version?

```
default_language_version: python3.7
```

iredmail:
My bad, Black require py3.6 to run.

cclauss:
black does not run on legacy Python.  The precommit Python is installed in its own virtualenv so this does not depend on or mess with the user's Python.

docs/conf.py
```diff
@@ -18,220 +18,214 @@
...
 
 # If your documentation needs a minimal Sphinx version, state it here.
-#needs_sphinx = '1.0'
+# needs_sphinx = '1.0'
 
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ['sphinx.ext.autodoc']
+extensions = ["sphinx.ext.autodoc"]
```
anandology: I'm really worried about making changes to the files like this just for pep-8 purity. IMHO, that is not a file created by us and I think we should leave it as it was generated.

experimental/untwisted.py
```diff
@@ -82,29 +93,30 @@ def publish(self, text):
...
 
 if __name__ == "__main__":
     mfeed = JSFeed()
 
-    urls = (
-      '/', 'view',
-      '/js', 'js',
-      '/send', 'send'
-    )
+    urls = ("/", "view", "/js", "js", "/send", "send")
```
anandology: This is not acceptable.

docs/conf.py
iredmail: Just replace single quotes by double quotes. I think it’s totally fine.

experimental/untwisted.py
iredmail: Agree. We lose readability for this (urls) reformat.

iredmail:
I’m on the opposite side: the places or new lines make it easier to read.

tests/test_application.py
```diff
@@ -24,242 +24,253 @@ def GET(self):
...
 
     def testUppercaseMethods(self):
         urls = ("/", "hello")
         app = web.application(urls, locals())
+
         class hello:
-            def GET(self): return "hello"
-            def internal(self): return "secret"
+            def GET(self):
+                return "hello"
```
anandology: Looks like the `internal` function disappeared. Bug in Black?

cclauss: It just moved down...  More like a bug in __diff__.

iredmail: The diff is correct.

cclauss:
In this PR, we make a lot of diff changes but on all future PRs, diffs become much easier to read because all contributors format code in exactly the same way.

tools/markdown.py
```diff
@@ -19,45 +19,49 @@
...
+__version__ = "1.0.1-2"  # port of 1.0.1
 __license__ = "GNU GPL 2"
 __author__ = [
-  'John Gruber <http://daringfireball.net/>',
-  'Tollef Fog Heen <tfheen@err.no>',
-  'Aaron Swartz <me@aaronsw.com>'
+    "John Gruber <http://daringfireball.net/>",
+    "Tollef Fog Heen <tfheen@err.no>",
+    "Aaron Swartz <me@aaronsw.com>",
 ]
```
anandology: Again, this is a external file included in web.py and it would be better to exclude this from pep-8.

iredmail: Since we maintain its copy instead of syncing it from upstream frequently, it's better format it too.

web/browser.py
```diff
@@ -105,66 +103,113 @@ def open(self, url, data=None, headers={}):
...
         e = e or self.get_soup()
-        return ''.join([htmlunquote(c) for c in e.recursiveChildGenerator()
-                       if isinstance(c, text_type)])
+        return "".join(
+            [
+                htmlunquote(c)
+                for c in e.recursiveChildGenerator()
+                if isinstance(c, text_type)
+            ]
+        )
```
anandology: This is a case where i find the initial version is more readable.

web/db.py
```diff
@@ -50,7 +50,7 @@
     "DB",
 ]
 
-TOKEN = '[ \\f\\t]*(\\\\\\r?\\n[ \\f\\t]*)*(#[^\\r\\n]*)?(((\\d+[jJ]|((\\d+\\.\\d*|\\.\\d+)([eE][-+]?\\d+)?|\\d+[eE][-+]?\\d+)[jJ])|((\\d+\\.\\d*|\\.\\d+)([eE][-+]?\\d+)?|\\d+[eE][-+]?\\d+)|(0[xX][\\da-fA-F]+[lL]?|0[bB][01]+[lL]?|(0[oO][0-7]+)|(0[0-7]*)[lL]?|[1-9]\\d*[lL]?))|((\\*\\*=?|>>=?|<<=?|<>|!=|//=?|[+\\-*/%&|^=<>]=?|~)|[][(){}]|(\\r?\\n|[:;.,`@]))|([uUbB]?[rR]?\'[^\\n\'\\\\]*(?:\\\\.[^\\n\'\\\\]*)*\'|[uUbB]?[rR]?"[^\\n"\\\\]*(?:\\\\.[^\\n"\\\\]*)*")|[a-zA-Z_]\\w*)'
+TOKEN = "[ \\f\\t]*(\\\\\\r?\\n[ \\f\\t]*)*(#[^\\r\\n]*)?(((\\d+[jJ]|((\\d+\\.\\d*|\\.\\d+)([eE][-+]?\\d+)?|\\d+[eE][-+]?\\d+)[jJ])|((\\d+\\.\\d*|\\.\\d+)([eE][-+]?\\d+)?|\\d+[eE][-+]?\\d+)|(0[xX][\\da-fA-F]+[lL]?|0[bB][01]+[lL]?|(0[oO][0-7]+)|(0[0-7]*)[lL]?|[1-9]\\d*[lL]?))|((\\*\\*=?|>>=?|<<=?|<>|!=|//=?|[+\\-*/%&|^=<>]=?|~)|[][(){}]|(\\r?\\n|[:;.,`@]))|([uUbB]?[rR]?'[^\\n'\\\\]*(?:\\\\.[^\\n'\\\\]*)*'|[uUbB]?[rR]?\"[^\\n\"\\\\]*(?:\\\\.[^\\n\"\\\\]*)*\")|[a-zA-Z_]\\w*)"
```
anandology: I'm kind of scared to make changes to this regular expression.

web/browser.py
iredmail: It could be more compact, but i'm ok with this new format, more (reasonable) space and more readable.

web/db.py
iredmail: It correctly handles the nested quotes.

iredmail:
> Can we exclude docs/conf.py and web/markdown.py from Black?

This should be fine. @cclauss could you update the PR?

anandology:
I'm okay (or can live with) with most of the changes, except the following:
- Changing the structure of `urls`
- Making changes to files that are generated or included from external sources (for example: `docs/conf.py` and `web/markdown.py`

iredmail:
Black is `uncompromising`, we have another choice if Black is not acceptable: [Google yapf](https://github.com/google/yapf) which offers more customizations.

cclauss:
Please wrap the URLs in __# fmt: off__ and __# fmt: on__

iredmail:
Hi @cclauss 

Could you help add `# fmt: off` for all `urls = ()`, and exclude `docs/conf.py`, `web/markdown.py`, then merge this PR?

cclauss:
I will deal with the excludes if you could please deal with the URLs.  I do not yet know where they all live.

iredmail:
No problem, but stupid question: how can i work based on existing changes in this PR? ^_*

cclauss:
Create a separate repo and once that repo has landed, I will rebase on master.

iredmail:
Is this correct?

- create new repo `iredmail/webpy`
- run `black` and submit changes

or create a new one under `https://github.com/webpy/`? Sorry i'm not so good at git/github.

btw, most `urls = ()` are defined in 3 files under `tests/`.

cclauss:
Sorry.  We do not need a new repo.  We just need a new PR on this repo.  The only changes in that PR should be the black comments.

iredmail:
Done: https://github.com/webpy/webpy/pull/524

PR #524, opened 2019-09-15
Add comment line to bypass Black format if URLs contain multiple patterns.

iredmail:
Add comment line `# fmt: off` for `urls = ()` which contain multiple patterns to bypass Black formatter.

### DIFF
--- a/experimental/pwt.py
+++ b/experimental/pwt.py
@@ -1,6 +1,8 @@
 from __future__ import print_function
 import web
 import simplejson, sudo
+
+# fmt: off
 urls = (
     '/sudo', 'sudoku',
     '/length', 'length',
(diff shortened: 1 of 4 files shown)


cclauss:
Thanks but, don’t you need to balance these with __# fmt: on__ or do I misunderstand.

iredmail:
For those `urls = ()` with only one url pattern like this:

```
urls = (
    '<pattern>', 'class-or-func'
)
```

it will be formatted to:

```
urls = ('<pattern>', 'class-or-func')
```

It's short and no readability issue at all, so we will let Black reformat it.

iredmail:
Just added the missing `# fmt: on`. Sorry about this stupid mistake.

iredmail:
Feel free to merge it and rebase your PR. :)

PR #536, opened 2019-09-17
pytest --doctest-modules .

cclauss:
Enable pytest to start running doctests on all supported versions of CPython.  There are two files run in __--ignore__ mode and the tests are run in __|| true__ to enable others to contribute fixes.

Remove vendoring-in of __markdown__.  Add it to both __setup.py__ and __requirements.txt__

Remove __simplejson__ in favor of Standard Library json.

Starting to sort the imports in an [__isort__](https://timothycrosley.github.io/isort/)-like way.  Should we do more of that.  It does aid readability.


iredmail:
@cclauss I did some research with Google to figure out how to make doctest working with py2+py3, i don't like any solutions i found.

How about this proposal:

- test only py2 for upcoming stable release - `0.40`
- drop py2 support and switch web.py to py3 completely
- make sure all doctests (and unit tests covered by pytest) working with py3

cclauss:
We should continue the path that we are already on... A single codebase that passes _all the same tests on BOTH_ Py2 and Py3.  When we are making modifications in this PR, we need to ensure that we still testing what we tested before those modifications.  When we drop legacy Python, it will be super-simple do a global-find-and-replace on __# doctest: +ALLOW_UNICODE__ to get rid of the ugly stuff.

anandology:
I think we should move away from doctests completely. It is very hard to
test stuff consistently with python 2 and python 3 using doctest. I've
started some work long time back to move all the tests to pytest and there
were some contributions too in that branch.

https://github.com/webpy/webpy/tree/pytest

I think it would be good if we complete that and merge it back to the
master.

PR #537, opened 2019-09-17
Fix DiskStore concurrency issue.

### LINKED ISSUE #83: EOFError when loading empty session files

### LINKED ISSUE #182: Race condition in setitem() of session.DiskStore

### LINKED ISSUE #191: session.DiskStore concurrency issue

### LINKED ISSUE #289: DiskStore session pickle load error caused by read and write conflict
it some times raise exceptions.EOFError

```
/usr/local/lib/python2.7/dist-packages/web/session.py in decode
        """encodes session dict as a string"""
        pickled = pickle.dumps(session_dict)
        return base64.encodestring(pickled)
    def decode(self, session_data):
        """decodes the data to get back the session dict """
        pickled = base64.decodestring(session_data)
>        return pickle.loads(pickled) 
```

I think it due to read conflict with write.. and I use FileLock from lockfile try to solve this..
Is this the right thing to do?

```
class DiskStore(Store):
     ....
    def __getitem__(self, key):
        path = self._get_path(key)
        if os.path.exists(path): 
            with FileLock(path) as lock:
                pickled = open(path).read()
            return self.decode(pickled)
        else:
            raise KeyError, key

        path = self._get_path(key)
        pickled = self.encode(value)    
        try:
            with FileLock(path) as lock:
                f = open(path, 'w')
                try:
                    f.write(pickled)
                finally: 
                    f.close()
        except IOError, e:
            print(e)
```

skawouter:
Write the diskstore in one atomary action
    
This will prevent multiple threads from accessing the file and possibly
read it while it is still being written to.

Fixes #191
Fixes #182 
Fixes #83
Fixes #289

This is in essence the same as my pull request #239 from 2013 with some unittesting.

### DIFF
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -90,7 +90,7 @@ class DiskStoreTest(unittest.TestCase):
         def set_val():
             store["fail"] = "value"
 
-        for c in xrange(10):
+        for c in range(10):
             m = threading.Thread(target=set_val)
             m.start()
             try:
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -2,6 +2,7 @@ import unittest
 import web
 import tempfile
 import os
+import threading
 
 
 class SessionTest(unittest.TestCase):
@@ -79,6 +80,22 @@ class SessionTest(unittest.TestCase):
         b.open("/redirect")
         b.open("/session/request_token")
         self.assertEqual(b.data, b"123")
+
+class DiskStoreTest(unittest.TestCase):
+    def testStoreConcurrent(self):
+        dir = tempfile.mkdtemp()
+        store = web.session.DiskStore(dir)
+        def set_val():
+            store['fail'] = 'value'
+
+        for c in xrange(10):
+            m = threading.Thread(target=set_val)
+            m.start()
+            try:
+                value = store['fail']
+            except KeyError:
+                pass
+        self.assertEquals(value, 'value')
 
 
 class DBSessionTest(SessionTest):
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -81,21 +81,23 @@ class SessionTest(unittest.TestCase):
         b.open("/session/request_token")
         self.assertEqual(b.data, b"123")
 
+
 class DiskStoreTest(unittest.TestCase):
     def testStoreConcurrent(self):
         dir = tempfile.mkdtemp()
         store = web.session.DiskStore(dir)
+
         def set_val():
-            store['fail'] = 'value'
+            store["fail"] = "value"
 
         for c in xrange(10):
             m = threading.Thread(target=set_val)
             m.start()
             try:
-                value = store['fail']
+                value = store["fail"]
             except KeyError:
                 pass
-        self.assertEquals(value, 'value')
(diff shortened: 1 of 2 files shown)


cclauss:
Please run __black .__ on the codebase and resubmit.  Using the precommit hook automates this.

cclauss:
__xrange()__ was removed in Python 3.  https://travis-ci.org/webpy/webpy/jobs/586390548#L283

tests/test_session.py
```diff
@@ -81,6 +82,24 @@ def testRedirect(self):
...
+            store["fail"] = "value"
+
+        for c in range(10):
+            m = threading.Thread(target=set_val)
+            m.start()
+            try:
+                value = store["fail"]
+            except KeyError:
+                pass
+        self.assertEquals(value, "value")
```
cclauss: __assertEquals()__ (with a trailing '__s__') was removed in Python 3 so please use __assertEqual()__ instead.

shanshangzhiren:
我的python2.7.10x86运行在Windows10x64，server2012等环境下，当web.py==0.40到0.51时，同样出现这个问题，按解决方案修改后，还是同样问题，但当web.py==0.39及以下时，工作正常。后，我把web/session.py从0.39复制到0.40或者0.51下覆盖，解决了这个问题。

cclauss:
We can no longer provide support for Python 2 which hould be avoided because it reached its end of life on 1/1/2020 which was 819 days ago.

PR #560, opened 2019-09-27
Render README.md into our PyPI page

cclauss:
https://pypi.org/project/web.py/ just has one line of text but https://pypi.org/project/tap2junit has its [README.md](https://github.com/nodejs/tap2junit/blob/master/README.md) file rendered in with links an all.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,10 +4,10 @@ import os
 from setuptools import setup
 from web import __version__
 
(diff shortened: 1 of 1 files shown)


iredmail:
I guess it only works for new package pushed to pypi?

cclauss:
Correct.  It is run in the __twine__ process.

PR #561, opened 2019-09-27
Should we add The web.py Philosophy to README.md?

cclauss:
From http://webpy.org/philosophy

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,3 +27,21 @@ python3 setup.py install
 
 Note: `0.40` is the last release which supports Python 2. Future releases
 will drop support for Python 2.
+
+### Philosophy
+
+The web.py slogan is: “_Think about the ideal way to write a web app. Write the code to make it happen._”
+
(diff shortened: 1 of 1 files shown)


iredmail:
It's good to have it in README.md.

This part was wrote by original author, Aaron, i think we have to rewrite it somehow to make it better for current situation. for example, i don't think it's good to write `This is literally how __I__ developed web.py.` Also, make it more serious to introduce a product.

cclauss:
I actually like the text in Arron's voice...  It talks directly to the developer in a way that is unique.

I would be interested in the views of @anandology and @sbp who have lived more of the history.

iredmail:
Then maybe add something LIKE "Said by original developer/founder, Aaron, ..." and quote the text below?

iredmail:
Hi @cclauss 

I suggest we add only this line to the README.md, then close this PR:

```
_Think about the ideal way to write a web app. Write the code to make it happen._
```

PR #563, opened 2019-09-28
Don't rely on 'markdown' module.

iredmail:
web.py is a micro web framework, it should just do the web framework thing, and leave non-framework things to other third-party modules.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -1,5 +1,12 @@
 # web.py changelog
 
+## 2019-XX-XX 0.50
+
+* Removed function `web.safemarkdown`. if it's used in your application, you
+  can install the `Markdown` module from pypi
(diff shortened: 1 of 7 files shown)


ChangeLog.txt
```diff
@@ -1,5 +1,12 @@
 # web.py changelog
 
+## 2019-XX-XX 0.50
```
cclauss: Semantic versioning would say that this should be in v1.0, not before.

iredmail:
I added alternative solution in ChangeLog.txt, it’s very easy. 

The point is making web.py just doing what a web framework should do and remove unnecessary functions. 

The safemarkdown was using a internally shipped python module and not updated for years, it’s better let users use the actively maintained third party modules. 

I suggest we remove it in next release, not after 1.0.

PR #565, opened 2019-10-02
Drop from imp import reload for Python3 < 3.4

cclauss:
We do not support Python 3.0, 3.1, 3.2, or 3.3


iredmail:
Oops, my mistake.

PR #573, opened 2019-10-24
Support tuple/set in `_sqllist()` and `sqlquote()`.

iredmail:
- Support tuple/set in `_sqllist()` and `sqlquote()`.
- Remove imported by unused module name.


iredmail:
Strange that i can not update the content of first post. The full changes should be:

```
- Support tuple/set in `_sqllist()` and `sqlquote()`.
- Remove imported by unused module name.
```

Issue #574, opened 2019-11-01
TypeError: write() argument must be str, not bytes


kjmph:
web.py release: 0.40
Python release: 3.7.3 (Debian buster)

While porting a project from Python 2 to Python 3 (thanks for the 0.40 release!), I immediately encountered a bug in `web.input()`. Apparently, when a POST is sent by `requests` (as opposed to a web browser), web.py emits an error:

```TypeError: write() argument must be str, not bytes```

If debug is enabled, the default, then a stream of errors occur with the a similar error. See this issue for more information https://github.com/webpy/webpy/issues/448.

Sure enough, setting `web.config.debug = False` will prevent the stream of errors, yet the originating error still exists. (Also, I want error handling, I'll address that later).

The source of the error is in `cgi.py`, as it tries and copy the BytesIO object web.py created into a TemporaryFile that was created without `wb+`.

Here is the cgi.py write error:
https://github.com/python/cpython/blob/ef4ec6ed12d6c6200a85068f60483723298b6ff4/Lib/cgi.py#L704

Here is the web.py BytesIO creation:
https://github.com/webpy/webpy/blob/fa27cc510e39af9c32e015c8daf51023a6e80007/web/webapi.py#L443-L444

I am able fix this by monkey patching `cgi.py`, as the source of the error is `make_file`:
https://github.com/python/cpython/blob/ef4ec6ed12d6c6200a85068f60483723298b6ff4/Lib/cgi.py#L814-L842

Here is a test case:
```python
import cgi
import web
import json
import tempfile

urls = (
    '/', 'index'
)

class index:
    def POST(self):
        inputs = web.input()
        data = json.loads(web.data())
        print(inputs, data)
        return "Hello, world!"

def make_file(self):
    return tempfile.TemporaryFile("wb+")

# Comment out this line to reproduce the error                                                                                                                               
cgi.FieldStorage.make_file = make_file

if __name__ == "__main__":
    # Comment this line to see the stream of errors. Note, make_file
    # also needs to be the default, not patched, version.
    web.config.debug = False
    app = web.application(urls, globals())
    app.run()
```

You may test this with httpie:
```shell
$ http -v POST localhost:8080 a=b
```

One last comment, when monkey patching cgi.py, the debugging error page works again.

Is there anything else I can submit to help solve this?

cclauss:
Python 2 makes no destiction between bytes and str but Python 3 strictly enforces that distinction:
```
$ python2 -c "print(bytes == str)"  # True
[...]

PR #575, opened 2019-11-02
Subclass cgi.FieldStorage to fix POST errors.

kjmph:
Subclass cgi.FieldStorage, as read_binary expects fp to return
bytes. If the headers do not contain a content-disposition with a
filename, cgi.FieldStorage's make_file will create a TemporaryFile
with `w+` flags. The write to that temporary file will fail, due to
incorrect encoding in Python 3.

See conversation in GitHub Issues:

https://github.com/webpy/webpy/issues/574

### DIFF
--- a/web/webapi.py
+++ b/web/webapi.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 import cgi
 import pprint
 import sys
+import tempfile
 from io import BytesIO
 
 from .py3helpers import PY2, text_type, urljoin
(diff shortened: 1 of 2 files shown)


web/webapi.py
```diff
@@ -401,7 +401,11 @@ class cgiFieldStorage(cgi.FieldStorage):
     to incorrect encoding in Python 3.
     """
 
-    def make_file(self):
+    def make_file(self, binary=None):
```
cclauss: If it is a flag then it is __True__ or __False__.

kjmph: Okay, I copied the style from Python 2's cgi. Yet, I can change the default value to `True`, as that follows what we are doing in the body of the method.

cclauss: Oh.  OK.  Then leave it as is.

anandology:
@iredmail and @cclauss do you see any issues with this?

iredmail:
LGTM :)

PR #579, opened 2019-11-20
Update templating.rst

rajesh-esaver:
Bullet numbers order is not correct, updated them.


iredmail:
Thanks for the contribution. Documents should go to another repo: https://github.com/webpy/webpy.github.com

PR #581, opened 2019-11-21
Fixed: base64.{encode,decode}string -> base64.{encode,decode}bytes.

iredmail:


cclauss:
The best practice is to [__use-feature-detection-instead-of-version-detection__](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection).

iredmail:
Thanks for sharing. I prefer version detection at this stage since we’re going to drop py2 support soon, it will be very easy to find py2 code by searching “PY2”. 

Feel free to switch to feature detection if that bothers you. :)

PR #594, opened 2020-02-04
Extracted Functionality from send() method in utils.py

ChrisAHolland:
The send() method in utils.py contained a lot of functionality that could be extracted to their own methods. Including:

- Setting up connections with AWS email server
- Setting up connections with an SMTP server

By extracting this functionality the send() method is reduced from 54 lines to 14, and easier to understand. Plus, if needed in the future, other methods could use the newly created methods to create connections with the email servers and send mail.

### DIFF
--- a/web/utils.py
+++ b/web/utils.py
@@ -1568,6 +1568,11 @@ class _EmailMessage:
         self.headers = {}
 
     def send(self):
+        try:
+            from . import webapi
+        except ImportError:
+            webapi = Storage(config=Storage())
+
         if webapi.config.get("smtp_server"):
             self.sendWithSMTP()
         elif webapi.config.get("email_engine") == "aws":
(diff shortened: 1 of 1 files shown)


iredmail:
I'm not sure whether it's necessary to refactor, because the class `__EmailMessage` is not expected to export.

Few modifications required in your PR:

- please rename the methods from `sendWithXXX` to `send_with_xxx`? For example: `sendWithAWS` -> `send_with_aws`.
- Better move below 2 lines outside of the `sendWithXXX` methods, and add a new function argument `message_text` instead.

```
self.prepare_message()
message_text = self.message.as_string()

def sendWithAWS(self, message_text):
    ...

def sendWithSMTP(self, message_text):
   ...
```

ChrisAHolland:
Thanks for the feedback! I thought the refactoring may be beneficial in reducing the method complexity. I will make the changes you have suggested.

PR #596, opened 2020-02-13
Add MemoryStore with simple unittest.

iredmail:
Fixes #174


ChangeLog.txt
```diff
@@ -2,6 +2,9 @@
 
 ## 2020-XX-XX 0.50
 
+* New session store `MemoryStore`, used to save a session in memory.
+  Should be useful where there is limited fs writes on the disk, like
```
cclauss: Change `is` —> `are` and `on` —> `to`.

Or an alternative:
Could be used with flash memory to minimize fs writes to disk.

PR #602, opened 2020-03-06
Try to test pgdb.

iredmail:


iredmail:
The `pgdb` seems not same as the one available on pypi: https://pypi.org/project/pgdb/#history
Give up.

PR #603, opened 2020-03-06
Isort will sort our imports so we don't have to

cclauss:
https://github.com/timothycrosley/isort


iredmail:
Looks good to me. :)

Issue #604, opened 2020-03-06
PEP8: Remove wildcard imports like "from db import *"


cclauss:
These can be discovered via: `flake8 --select=F4 --statistics`
```
25    F401 'os' imported but unused
9     F403 'from .application import *' used; unable to detect undefined names
```

And can be fixed with https://www.asmeurer.com/removestar

iredmail:
Removing stars is much better, let’s do this.

anandology:
What will that be replaced with?

cclauss:
```python
from . import session  # noqa: F401
from .application import (  # noqa: F401
    application,
    auto_application,
    loadhook,
    subdomain_application,
    unloadhook,
)
from .db import database  # noqa: F401
from .http import changequery  # noqa: F401
from .utils import storage  # noqa: F401
from .webapi import (  # noqa: F401
    HTTPError,
    cookies,
    ctx,
    input,
    notfound,
    redirect,
    seeother,
    setcookie,
)
```

PR #605, opened 2020-03-06
PEP8: Remove wildcard imports like "from db import *"

cclauss:
Fixes: #604   This needs careful review because it might have put enough imports in place to pass the tests but other apps might be looking for other imports.

### DIFF
--- a/web/__init__.py
+++ b/web/__init__.py
@@ -27,6 +27,7 @@ from .utils import *
 from .webapi import *
 from .wsgi import *
 """
+from . import application  # noqa: F401
 
 __version__ = "0.40"
 __author__ = [
--- a/web/__init__.py
+++ b/web/__init__.py
@@ -3,6 +3,7 @@
 
 from __future__ import generators
 
+"""
 from . import (
     db,
     debugerror,
@@ -25,6 +26,7 @@ from .net import *
 from .utils import *
 from .webapi import *
 from .wsgi import *
+"""
 
 __version__ = "0.40"
 __author__ = [
--- a/web/__init__.py
+++ b/web/__init__.py
@@ -3,6 +3,13 @@
 
 from __future__ import generators
 
+from .application import subdomain_application  # noqa: F401
+from .application import application, auto_application  # noqa: F401
+from .db import database  # noqa: F401
+from .http import changequery  # noqa: F401
+from .utils import storage  # noqa: F401
+from .webapi import HTTPError, ctx, input, seeother, setcookie  # noqa: F401
+
 """
 from . import (
     db,
@@ -27,7 +34,6 @@ from .utils import *
 from .webapi import *
 from .wsgi import *
 """
-from .application import application  # noqa: F401
 
 __version__ = "0.40"
 __author__ = [
--- a/web/__init__.py
+++ b/web/__init__.py
@@ -3,8 +3,9 @@
 
 from __future__ import generators
 
-from .application import subdomain_application  # noqa: F401
-from .application import application, auto_application  # noqa: F401
+from . import session
+from .application import application, auto_application, loadhook  # noqa: F401
+from .application import subdomain_application, unloadhook  # noqa: F401
 from .db import database  # noqa: F401
 from .http import changequery  # noqa: F401
 from .utils import storage  # noqa: F401
--- a/web/__init__.py
+++ b/web/__init__.py
@@ -1,32 +1,7 @@
 #!/usr/bin/env python
 """web.py: makes web apps (http://webpy.org)"""
 
-from __future__ import generators
-
-from . import session  # noqa: F401
-from .application import (  # noqa: F401
-    application,
-    auto_application,
-    loadhook,
-    subdomain_application,
-    unloadhook,
-)
-from .db import database  # noqa: F401
-from .http import changequery  # noqa: F401
-from .utils import storage  # noqa: F401
-from .webapi import (  # noqa: F401
-    HTTPError,
-    cookies,
-    ctx,
-    input,
-    notfound,
-    redirect,
-    seeother,
-    setcookie,
-)
-
-"""
-from . import (
+from . import (  # noqa: F401
     db,
     debugerror,
     form,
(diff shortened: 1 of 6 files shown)


anandology:
This is going to break lot of things. Here is some code that I'm currently working on that will stop working after this change. 

```
query.py:            q += web.reparam("\nLIMIT $limit", {"limit": self.limit})
...
query.py:        q += "\nWHERE " + web.SQLQuery.join(self.wheres, " AND ")
```

Every module in webpy has `__all__` defined and all those are imported. I don't really understand the real benefit of removing import *. I know that not considered as a good practice in general and I don't mind taking exception in this case. It hurts readability a bit, but given that we have __all__ defined in every module, it is not that bad. 

Finally, I don't think the benefit that we are getting out of this exercise, justifies the risk that is involved with it.

iredmail:
- I agree to remove `from ... import *`, it definitely improves the readability (explicit vs implicit).
- `__all__` in each module is not same thing as the `import *`.

But @anandology's code sample does show the concern. Many sample webpy code on website or tutorials use `web.reparam()` (which should be `web.db.reparam()`) and similar namespaces due to the `from ... import *` thing. I didn't realize this before.

To keep backward compatibility, we have to keep the stars in just `web/__init__.py` file, but not in any other `web/*.py`.

anandology:
On Sat, Mar 7, 2020 at 9:26 AM iRedMail <notifications@github.com> wrote:

>
>    - I agree to remove from ... import *, it definitely improves the
>    readability (explicit vs implicit).
>    - __all__ in each module is not same thing as the import *.
>
> But @anandology <https://github.com/anandology>'s code sample does show
> the concern. Many sample webpy code on website or tutorials use
> web.reparam() (which should be web.db.reparam()) and similar namespaces
> due to the from ... import * thing. I didn't realize this before.
>
> To keep backward compatibility, we have to keep the stars in just
> web/__init__.py file, but not in any other web/*.py.
>
We do not have `import *`  in any `web/*.py` except `__init__.py`.

iredmail:
@cclauss let's keep the stars in `web/__init__.py` for backward compatibility but update our sample code and tutorials to avoid this.

cclauss:
What about this solution...  What if we take the content from each of the `__all__` statements and copy them into explicit imports in `__init__.py`.  That way we are explicitly living up to the same api contract but with all imported names/locations are in one place for easy reference.

iredmail:
> What about this solution... What if we take the content from each of the `__all__` statements and copy them into explicit imports in `__init__.py`.

Agree. It's a balance.

The replaced `import` code in `web/__init__.py` may look ugly and `from ... import *` is much cleaner, but it's better to get explicit imports and stop encouraging users to use namespace like `web.XXX` (which should be `web.Mod.XXX`).

anandology:
On Sat, Mar 7, 2020 at 9:47 AM Christian Clauss <notifications@github.com>
wrote:

> What about this solution... What if we take the content from each of the
> __all__ statements and copy them into explicit imports in __init__.py.
> That way we are explicitly living up to the same api contract but with all
> imported names/locations are in one place for easy reference.
>
Lets not make changes just for the sake of purity. Adding __all__ to all
modules and using import * in __init__.py is a well known pattern. Here is
a recent example:

https://github.com/dabeaz/curio/blob/master/curio/__init__.py

iredmail:
@anandology It's not about `purity` at all. Also, other projects use this bad practice doesn't make it a good practice.

See discussion at https://www.asmeurer.com/removestar "Why is import * so bad?"

anandology:
On Sat, Mar 7, 2020 at 10:08 AM iRedMail <notifications@github.com> wrote:

> @anandology <https://github.com/anandology> It's not about purity at all.
> Also, other projects use this bad practice doesn't make it a good practice.
>
> See discussion at https://www.asmeurer.com/removestar "Why is import * so
> bad?"
>
import * is bad as a public API, but it is fine as an implementation
details.

If I had to do "from web import *" to use webpy, i would be worried, but
not with what we have now.

iredmail:
I would suggest we go @cclauss's proposal https://github.com/webpy/webpy/pull/605#issuecomment-596044644, let's change it and not import any NEW ones in `__init__.py`. It keeps perfect backward compatibility, but also avoid bad practice.

cclauss:
It is a long list of imports but it is explicit.

anandology:
I fail to understand what is the benefit of doing this. I don't think having `import *` in `__init__.py` is a bad idea, it is an implementation detail. I prefer to continue the same instead of enumerating every exposed name in __init__.py.

web/__init__.py
```diff
@@ -16,15 +14,160 @@
...
-from .application import *
-from .db import *
-from .debugerror import *
-from .http import *
-from .httpserver import *
-from .net import *
-from .utils import *
-from .webapi import *
-from .wsgi import *
+from .application import (  # noqa: F401
```
iredmail: Please add some (brief) comment lines in `web/__init__.py` to mention why we do this change, and emphasize not to introduce any new functions in the `from .XX import (...)`. For example:

> In web.py-0.40 and earlier releases, we use `from .XX import *` to import all
> (exported) functions from sub-modules to `web.*` namespace. For example,
> `web.reparam()` (should be `web.db.reparam()`), `web.safestr()` (should be
> `web.utils.safestr()`). We decide to stop this bad practice and encourage
> users to use clear namespaces (use `web.XX.YY` instead of `web.YY`).
>
> These explicit imports are kept for backward-compatibility, do not import
> any namespace under `web.` for new functions.
>
> FYI: https://www.asmeurer.com/removestar/

Just an example, please update/adjust with better / clearer explanation.

anandology:
> Please add some (brief) comment lines in web/__init__.py to mention why
> we do this change, and emphasize not to introduce any new functions in the from
> .XX import (...). For example:
>
> In web.py-0.40 and earlier releases, we use from .XX import * to import
> all
> (exported) functions from sub-modules to web.* namespace. For example,
> web.reparam() (should be web.db.reparam()), web.safestr() (should be
> web.utils.safestr()). We decide to stop this bad practice and encourage
> users to use clear namespaces (use web.XX.YY instead of web.YY).
>
> First, I don't agree that it is a bad practice and I don't remember that
we decided to to encourage users to use namespaces.

> These explicit imports are kept for backward-compatibility, do not import
> any namespace under web. for new functions.
>
> I don't agree with this.

Flat is better than nested. Putting all functionality in a single namespace
was a conscious decision taken from the very early days. I suggest to
continue that.

iredmail:
- 10 years ago we didn't really care about the stars, but it is changed nowadays.
- This change keeps perfect backward compatibility, no application will be broken after upgraded to upcoming new release with this change.

We're now in interesting situation: @cclause and me tend to change it, but you want to keep current state. What are we gonna do now? change or stay? 😄

iredmail:
@anandology by the way, did you check https://www.asmeurer.com/removestar/ ?

anandology:
On Mon, Mar 16, 2020 at 9:19 AM iRedMail <notifications@github.com> wrote:

> @anandology <https://github.com/anandology> by the way, did you check
> https://www.asmeurer.com/removestar/ ?
>

Yes, did you notice this from that page?

"It is considered acceptable when working interactively at a python prompt,
or in __init__.py files (removestar skips __init__.py files by default)."

cclauss:
Closing.  @anandology is right.  If it is not broken don’t fix it.  The wildcard imports are an implementation detail.

PR #610, opened 2020-03-14
ALLOWED_AST_NODES: readability and maintainability

cclauss:


iredmail:
A better change would be excluding unsafe nodes instead of whitelisting safe ones.

cclauss:
> I’m not sure why this change is better than a list.

1. 82 lines removed from the file.
2. Read all content without scrolling.
3. Easier to spot if the item you want is present.
4. Easier to spot the duplicates.

> A better change would be excluding unsafe nodes...

I disagree.  For spotting security concerns, it is better to be able to scan all the enabled items.  For example, many users might be scared to see `Exec` on this list.  We should not hide the fact that we enabling it that from them.

anandology:
I prefer to keep the list of nodes one per line. Combining all of them into a single line makes it too hard to comprehend. If you would like to reduce the number of lines, try grouping related nodes into a single line, but I would still like to keep it as multiple lines.

anandology:
> A better change would be excluding unsafe nodes instead of whitelisting safe ones.

No, it was a design decision to whitelist the safe nodes. If we do the other way, the system becomes insecure the moment an unsafe node is added to the next version of Python.

Issue #612, opened 2020-03-14
Python3: web.template.render(TEMPLATES_DIR, base=layout) does not work


christophe-dupriez:
I am trying to switch my applications to Python3. If I use a "base template", I always end with the following message (whatever workaround I tried). Without "base template", no problem (for now in very reduced test application). 

templates/layout.html:
```
<!doctype html>
<head>
	<meta charset="utf-8">
</head>
<body>
    <h1>A title!</h1>
    $:content
</body>
</html>

```
Error messages with parameter "base=layout" added to template.render:
```
/usr/bin/python3.5 /home/christophe/PycharmProjects/PandLoRa/test.py
http://0.0.0.0:8080/
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/web.py-0.40-py3.5.egg/web/application.py", line 290, in process
    return self.handle()
  File "/usr/local/lib/python3.5/dist-packages/web.py-0.40-py3.5.egg/web/application.py", line 281, in handle
    return self._delegate(fn, self.fvars, args)
  File "/usr/local/lib/python3.5/dist-packages/web.py-0.40-py3.5.egg/web/application.py", line 531, in _delegate
    return handle_class(cls)
  File "/usr/local/lib/python3.5/dist-packages/web.py-0.40-py3.5.egg/web/application.py", line 509, in handle_class
    return tocall(*args)
  File "/home/christophe/PycharmProjects/PandLoRa/test.py", line 43, in GET
    return render.index(True,"dupriez@destin.be","192.168.0.1")
  File "/usr/local/lib/python3.5/dist-packages/web.py-0.40-py3.5.egg/web/template.py", line 1140, in template
    return self._base(t(*a, **kw))
  File "/usr/local/lib/python3.5/dist-packages/web.py-0.40-py3.5.egg/web/template.py", line 987, in __call__
    return BaseTemplate.__call__(self, *a, **kw)
  File "/usr/local/lib/python3.5/dist-packages/web.py-0.40-py3.5.egg/web/template.py", line 898, in __call__
    return self.t(*a, **kw)
TypeError: __template__() takes 0 positional arguments but 1 was given

Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/web.py-0.40-py3.5.egg/web/application.py", line 290, in process
    return self.handle()
  File "/usr/local/lib/python3.5/dist-packages/web.py-0.40-py3.5.egg/web/application.py", line 281, in handle
[...]

PR #617, opened 2020-03-20
Clarify TypeError if layout.html does not start with '$def with (content)'

cclauss:
Partial fix for #612 but we also need a test that shows that the proper exception will be raised.

### DIFF
--- a/web/template.py
+++ b/web/template.py
@@ -895,7 +895,11 @@ class BaseTemplate:
 
     def __call__(self, *a, **kw):
         __hidetraceback__ = True
-        return self.t(*a, **kw)
+        try:
+            return self.t(*a, **kw)
+        except TypeError as e:
+            print("layout.html must have '$def with (content)' near the beginning.")
+            raise e
 
     def make_env(self, globals, builtins):
         return dict(


cclauss:
We still raise a TypeError so this is not a warning but the message is more helpful.

I do not know how to add the test.

iredmail:
Hmm, I replied too quick. Some concerns of this change:

- the variable name “content” could be any name. The “print()” line may implicitly advice users to use just “content”.

- Seems this change checks all template files, not just layout file. So the text in “print()” may mislead user to check just layout.html. 

- Will “TypeError” also catch other errors?

cclauss:
The try / except block will __only__ catch TypeErrors.

iredmail:
This PR is not correct.

- The `except` is added in wrong place.
- Other errors may raise `TypeError` exception too, so the text in `print()` is incorrect.

I will send a new PR for this.

iredmail:
There're at least 3 different errors might be raised while rendering a template file, and it could happen to any template file, not just the layout.html, so the text in your `print()` is imprecise.

Proposal:

- Update [templator document](https://webpy.org/docs/0.3/templetor), list all possible errors and the solution for each error.
- Catch `TypeError` and `NameError` in your PR, print text with a link to the document. For example:

```
        try:
             return self.t(*a, **kw)
         except (TypeError, NameError) as e:
             print("Check web.py website to check possible errors and how to fix it: https://webpy.org/docs/0.3/templetor#errors
             raise e
```

#### `TypeError: __template__() got an unexpected keyword argument 'myvar'`

The `render` calls the template file with unexpected keyword argument `myvar`.

Template file:
```
$def with (var1, var2)

My content:
$:var1
$:var2
```

Python code.

```
import web
render = web.template.render()

# WRONG. `myvar` should be removed.
print(render.my_template_file(var1=..., var2=..., myvar=...))

# CORRECT.
print(render.my_template_file(var1=..., var2=...))
```

#### `TypeError: __template__() takes 0 positional arguments but 1 was given`

Template file doesn't have `$def with (...)` line (which indicates it doesn't use any variables), but 1 variable name is passed to the template while calling the `render` function.

Template file:
```
My sample content here without using any variable.
```

Python code:

```
# WRONG.
print(render.my_template_file(myvar=...))

# CORRECT.
print(render.my_template_file())
```

#### `TypeError: __template__() missing 1 required positional argument: 'myvar'`

Template file uses variable name `myvar`, but it's not passed while calling the `render`.

Template file:

```
$def with (myvar)

Render the value:
$:myvar
```

Python code.
```
# WRONG.
print(render.my_template_file())

# CORRECT
print(render.my_template_file(myvar=...))
```

cclauss:
Closing.  Please create a PR to replace it.

iredmail:
@cclauss could you help create a new PR? Just change the text in the “print()” should be enough. :)

cclauss:
I would encourage you to do...
```python
try:
    ABC
except NameError as e:
    print("Message about NameErrors")
    raise e
except TypeError as e:
    print("Message about TypeErrors")
    raise e
```

cclauss:
If you create the PRthen I can review it.

PR #618, opened 2020-03-21
Print message with link to possible errors/solutions while failed to …

iredmail:
Print message with link to possible errors/solutions while failed to render template files with the builtin templator.

### DIFF
--- a/web/template.py
+++ b/web/template.py
@@ -898,8 +898,10 @@ class BaseTemplate:
         try:
             return self.t(*a, **kw)
         except TypeError as e:
-            print("Check web.py website to check possible errors and how to "
-                  "fix it: https://webpy.org/docs/0.3/templetor#errors")
+            print(
+                "Check web.py website to check possible errors and how to "
+                "fix it: https://webpy.org/docs/0.3/templetor#errors"
+            )
             raise e
 
     def make_env(self, globals, builtins):
--- a/web/template.py
+++ b/web/template.py
@@ -895,7 +895,12 @@ class BaseTemplate:
 
     def __call__(self, *a, **kw):
         __hidetraceback__ = True
-        return self.t(*a, **kw)
+        try:
+            return self.t(*a, **kw)
+        except TypeError as e:
+            print("Check web.py website to check possible errors and how to "
+                  "fix it: https://webpy.org/docs/0.3/templetor#errors")
+            raise e
 
     def make_env(self, globals, builtins):
         return dict(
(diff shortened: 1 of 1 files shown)


iredmail:
`NameError` is kind of easy to locate and fix, so not covered in this PR.

web/template.py
```diff
@@ -895,7 +895,14 @@ def _compile(self, code):
 
     def __call__(self, *a, **kw):
         __hidetraceback__ = True
-        return self.t(*a, **kw)
+        try:
+            return self.t(*a, **kw)
+        except TypeError as e:
+            print(
```
cclauss:
```suggestion
            print("To fix this error, please see https://webpy.org/docs/0.3/templetor#errors")
```

web/template.py
```diff
@@ -895,7 +895,13 @@ def _compile(self, code):
...
+                "To fix this error, please see https://webpy.org/docs/0.3/templetor#errors"
+            )
+            raise e
```
anandology: I see couple of issues with this:
- print is going to stdout instead of stderr. 
- putting links in the code and error messages is quite risky as the links become stale and we wouldn't even know. Please keep a note to fix this link when we release 1.0.

@iredmail

cclauss: We can add a test case to ensure the URL is still valid.

anandology:
On Sat, Mar 21, 2020 at 1:26 PM Christian Clauss <notifications@github.com>
wrote:

> *@cclauss* commented on this pull request.
> ------------------------------
>
> In web/template.py
> <https://github.com/webpy/webpy/pull/618#discussion_r395971869>:
>
> > @@ -895,7 +895,13 @@ def _compile(self, code):
>
>      def __call__(self, *a, **kw):
>          __hidetraceback__ = True
> -        return self.t(*a, **kw)
> +        try:
> +            return self.t(*a, **kw)
> +        except TypeError as e:
> +            print(
> +                "To fix this error, please see https://webpy.org/docs/0.3/templetor#errors"
> +            )
> +            raise e
>
> We can add a test case to ensure the URL is still valid.
>
You can only verify that the URL exists. How do you verify that the
contents of the page are valid? Also, giving a link to 0.3 docs when you
are working on version 0.50 is confusing.

iredmail:
> Also, giving a link to 0.3 docs when you are working on version 0.50 is confusing.

I'd like to point to a up to date document, but you know we don't have documents for 0.40, 0.50 and upcoming 1.0 ...

PR #622, opened 2020-03-23
Fix shebang lines to run Python3

cclauss:


setup.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import os
```
iredmail: Please remove the py2 related info in `classifiers=[]`.

Issue #624, opened 2020-03-25
cheroot as optional dependency


stroeder:
Is cheroot really required? I run my web apps via uwsgi without cheroot just fine. Thus I'd like to avoid having to install cheroot and its dependencies.

iredmail:
cheroot is not required if you run your web.py application with other wsgi/cgi/fastcgi server like uwsgi, gunicorn, etc. But it's required if you run it with command `python your-app.py` directly, it's used as builtin wsgi server.

I personally prefer to remove cheroot, then no any dependent external packages.

iredmail:
hi @cclauss @anandology 

What's your opinion about removing `cheroot`?

- No one run `pythonX app.py` (and/or proxy requests in Nginx) in production.
- We can document in `README.md` and ask users to install tools like `gunicorn` or `uwsgi` to run web.py app as wsgi application in development mode, also setup Nginx/Apache/... web servers to proxy the http requests.

cclauss:
I have no opinion on this topic.

iredmail:
> I have no opinion on this topic.

Does this mean either removing or keeping it is ok for you?

cclauss:
For me, things are OK the way they are.  I believe that there is a lot to be said for "_It works out of the box".

If others feel passionately and can justify why multiple projects should change the way they install web.pyand are willing to modify our example code to work simply then I would listen to their reasons.

PR #626, opened 2020-03-25
Install cheroot only if requested as [cheroot]. #624

stroeder:
Tested in virtual environment whether it has desired effect.


cclauss:
flake8 errors...
https://github.com/webpy/webpy/pull/626/checks?check_run_id=534405322#step:8:5

Is cheroot needed on other OSes?

stroeder:
> Is cheroot needed on other OSes?

I don't understand the question. AFAICS cheroot is pure-Python. I never used web.py on anything else than Linux though.

Issue #630, opened 2020-03-31
def with (exception_type, exception_value, frames) ^ SyntaxError: invalid syntax


regmiscacc:
Here is my code.py:
import web
render = web.template.render('templates/')

urls = (
  '/', 'index'
)

class index:
  def GET(self):
    return render.index()

if __name__ == "__main__":
 app = web.application(urls, globals())
 app.run()

And here is my templates/index.html
`<em>Hello</em>, world!`

When i run it, i getting this error:
**Traceback (most recent call last):
  File "C:\Python38\lib\site-packages\cheroot\server.py", line 1269, in communicate
    req.respond()
  File "C:\Python38\lib\site-packages\cheroot\server.py", line 1071, in respond
    self.server.gateway(self).respond()
  File "C:\Python38\lib\site-packages\cheroot\wsgi.py", line 143, in respond
    response = self.req.server.wsgi_app(self.env, self.start_response)
  File "C:\Python38\lib\site-packages\web\httpserver.py", line 306, in __call__
    return self.app(environ, xstart_response)
  File "C:\Python38\lib\site-packages\web\httpserver.py", line 273, in __call__
    return self.app(environ, start_response)
  File "C:\Python38\lib\site-packages\web\application.py", line 329, in wsgi
    result = self.handle_with_processors()
  File "C:\Python38\lib\site-packages\web\application.py", line 300, in handle_with_processors
    return process(self.processors)
  File "C:\Python38\lib\site-packages\web\application.py", line 297, in process
    raise self.internalerror()
  File "C:\Python38\lib\site-packages\web\application.py", line 588, in internalerror
    return debugerror()
  File "C:\Python38\lib\site-packages\web\debugerror.py", line 337, in debugerror
    return web._InternalError(djangoerror())
  File "C:\Python38\lib\site-packages\web\debugerror.py", line 315, in djangoerror
    djangoerror_r = Template(djangoerror_t, filename=__file__, filter=websafe)
  File "C:\Python38\lib\site-packages\web\template.py", line 943, in __init__
    code = self.compile_template(text, filename)
  File "C:\Python38\lib\site-packages\web\template.py", line 1020, in compile_template
    compiled_code = compile(code, filename, "exec")
  File "C:\Python38\lib\site-packages\web\debugerror.py", line 8
    def with (exception_type, exception_value, frames)
        ^
SyntaxError: invalid syntax**

Why?

cclauss:
No semicolon at the end of the line???

[...]

PR #631, opened 2020-04-01
Fix escape string from causing syntax error (#630)

imbellish:


imbellish:
🤦‍♂

cclauss:
Clicking on the failed tests...
```
./web/debugerror.py:95:46: W605 invalid escape sequence '\-'
        strClassName = strClassName.replace(/\-/g, "\\-");
                                             ^
1     W605 invalid escape sequence '\-'
1
##[error]Process completed with exit code 1
```

imbellish:
@cclauss I couldn't find the original commit for this template, last I checked, in the OG django logs. And I think there's more work to do here to support python>=3.5 but not python <=2.7.

cclauss:
Hint:
$ `python`
```
>>> "my string".replace(/\-/g, "\\-")
  File "<string>", line 1
    "my string".replace(/\-/g, "\\-")
                        ^
SyntaxError: invalid syntax
```

teebee:
Why is https://github.com/webpy/webpy/blob/3f3fd36ed87d138359c66fc451d3fdbeb89071bd/web/debugerror.py#L96 needed at all? A dash is only special inside a [[...]](https://javascript.info/regexp-character-sets-and-ranges#escaping-in-). Am I missing something?

teebee:
It seems that `flake8` is used to check JavaScript code here ...

cclauss:
Hint again: You need to put strings in quotes.

iredmail:
@cclauss Do i remember correctly that this issue had been fixed in some other commit? Can we close this PR now?

cclauss:
I believe that this is fixed in the current releases. Can you please retest on v0.6x and let us know if we can close this issue?

PR #645, opened 2020-06-05
Fix incorrect str/bytes type of session data.

### LINKED ISSUE #644: example in tutorial on URL 'https://webpy.org/cookbook/sessions' is not working.
Hi, 
Refering to tutorial on URL https://webpy.org/cookbook/sessions 

the sample code is alright except that if implemented in a Postgres DB, the data-type of the data column must be bytea instead of text. Otherwise, the following error is reported during run-time. 

```
  File "/usr/lib/python3.6/base64.py", line 510, in _input_type_check
    m = memoryview(s)
TypeError: memoryview: a bytes-like object is required, not 'str'

The above exception was the direct cause of the following exception:

  File "/home/abhijit/.local/lib/python3.6/site-packages/web/application.py", line 288, in process
    return p(lambda: process(processors))
  File "/home/abhijit/.local/lib/python3.6/site-packages/web/session.py", line 102, in _processor
    self._load()
  File "/home/abhijit/.local/lib/python3.6/site-packages/web/session.py", line 120, in _load
    d = self.store[self.session_id]
  File "/home/abhijit/.local/lib/python3.6/site-packages/web/session.py", line 348, in __getitem__
    return self.decode(s.data)
  File "/home/abhijit/.local/lib/python3.6/site-packages/web/session.py", line 246, in decode
    pickled = decodebytes(session_data)
  File "/usr/lib/python3.6/base64.py", line 545, in decodebytes
    _input_type_check(s)
  File "/usr/lib/python3.6/base64.py", line 513, in _input_type_check
    raise TypeError(msg) from err
TypeError: expected bytes-like object, not str

127.0.0.1:54780 - - [22/May/2020 17:58:54] "HTTP/1.1 GET /count" - 500 Internal Server Error
```
In the interest of beginners(such as me) in the community, please update the document with the correct database table specification(likely as follows). I found it working without issues on Python3.6 with Postgresql 12. 

`create table sessions ( session_id char(128) primary key, atime timestamp default now(), data bytea) ;`

Thanks,
Abhijit

iredmail:
Fixes #644

Works with:
- MySQL column type `TEXT` and `VARBINARY`
- PostgreSQL column type `TEXT` and `bytea`

py3 code i used for testing:

```
import web

web.config.debug = False

urls = (
    "/count", "count",
)

app = web.application(urls, locals())

conn = web.database(
    dbn='mysql',    # Use `postgres` for PostgreSQL
    host='localhost',
    port=3306,    # Use `5432` for PostgreSQL
    db='<db-name>',
    user='<db-user>',
    pw='<db-password>',
)

session = web.session.Session(
    app=app,
    store=web.session.DBStore(conn, "sessions"),
    initializer={"count": 0},
)

class count:
    def GET(self):
        session.count += 1
        return str(session.count)

if __name__ == "__main__":
    app.run()
```

### DIFF
--- a/web/session.py
+++ b/web/session.py
@@ -240,9 +240,9 @@ class Store:
     def decode(self, session_data):
         """decodes the data to get back the session dict """
         if isinstance(session_data, str):
-            pickled = decodebytes(session_data.encode())
-        else:
-            pickled = decodebytes(session_data)
+            session_data = session_data.encode()
+
+        pickled = decodebytes(session_data)
         return pickle.loads(pickled)
 
 
@@ -347,7 +347,8 @@ class DBStore(Store):
             return self.decode(s.data)
 
     def __setitem__(self, key, value):
-        pickled = self.encode(value)
+        # Remove the leading `b` of bytes object.
+        pickled = self.encode(value).decode()
         now = datetime.datetime.now()
         if key in self:
             self.db.update(
--- a/web/session.py
+++ b/web/session.py
@@ -239,7 +239,10 @@ class Store:
 
     def decode(self, session_data):
         """decodes the data to get back the session dict """
-        pickled = decodebytes(session_data)
+        if isinstance(session_data, str):
+            pickled = decodebytes(session_data.encode())
+        else:
+            pickled = decodebytes(session_data)
         return pickle.loads(pickled)
 
 
--- a/web/session.py
+++ b/web/session.py
@@ -347,8 +347,10 @@ class DBStore(Store):
             return self.decode(s.data)
 
     def __setitem__(self, key, value):
-        # Remove the leading `b` of bytes object.
+        # Remove the leading `b` of bytes object (`b"..."`), otherwise encoded
+        # value is invalid base64 format.
         pickled = self.encode(value).decode()
+
         now = datetime.datetime.now()
         if key in self:
             self.db.update(


cclauss:
python2: >>> bytes == str  # --> True
python3: >>> bytes == str  # --> False

iredmail:
You're right, the fix is only applicable to py3 and tested locally. Out master branch is py3 only, so it should be safe to apply it.
For 0.5X branch, we need some more attention.

iredmail:
Thanks @cclauss 

I'd like to wait for abhijitnarvekar's reply (author of issue #644) before merging this PR.

iredmail:
hmm, this fix works fine with MySQL column type `TEXT` and `VARBINARY`, but not `TEXT` for PGSQL. It raises error on Ubuntu 20.04 with PostgreSQL 12.1:

```
  File "/usr/local/lib/python3.8/dist-packages/web/application.py", line 288, in process
    return p(lambda: process(processors))
  File "/usr/local/lib/python3.8/dist-packages/web/session.py", line 102, in _processor
    self._load()
  File "/usr/local/lib/python3.8/dist-packages/web/session.py", line 120, in _load
    d = self.store[self.session_id]
  File "/usr/local/lib/python3.8/dist-packages/web/session.py", line 351, in __getitem__
    return self.decode(s.data)
  File "/usr/local/lib/python3.8/dist-packages/web/session.py", line 247, in decode
    pickled = decodebytes(session_data.encode())
  File "/usr/lib/python3.8/base64.py", line 546, in decodebytes
    return binascii.a2b_base64(s)
binascii.Error: Invalid base64-encoded string: number of data characters (277) cannot be 1 more than a multiple of 4
```

cclauss:
;-). Looks like we should add an automated test.

abhijitnarvekar:
Hello,

I had reported the issue and in the same post proposed a solution to the
issue. The adaption to MySql is also correct. However, most examples in the
tutorial are drawn from Postgresql. So i just discussed Postgresql.

Thanks,
Abhijit

iredmail:
Modifying sql column type is a solution, but we’d like to support TEXT type also. 

Did you help test the modification in this PR?

iredmail:
Dear @cclauss @abhijitnarvekar 

Please help test and review, did one more change and finally get it working with both MySQL and PostgreSQL. Sample Python3 code is attached in first comment in this PR.

iredmail:
There's one "issue" with this fix: Since the data is base64 encoded, there's a new line `\n` after ever 76 characters. Original one was a long string without `\n`.

web/session.py
```diff
@@ -347,7 +347,8 @@ def __getitem__(self, key):
             return self.decode(s.data)
 
     def __setitem__(self, key, value):
-        pickled = self.encode(value)
+        # Remove the leading `b` of bytes object.
```
cclauss: I think we can lose this comment because it is misleading to say that decoding bytes is merely removing the b from the front.

iredmail: The comment line was updated BEFORE your comment.
I  think it's necessary to leave some comment to help others who read the code to understand why we do `.encode(...).decode()` here. If the words are not good enough (English is not my first language), feel free to update it.

lujoamel4:
I'm thinking of the framework design.

Should webpy support both schema configurations? I think that only one way is better. 

As the Python Zen says: 
There should be one-- and preferably only one --obvious way to do it.

So, I think there are two options:
1. Change database table and docs in webpy.org 
2. Change code in session.py

I will be waiting for your comments.

Thanks!

iredmail:
@lujoamel4 

Dropping support for `TEXT/VARCHAR` will break existing webpy applications, so better support both in current stage.

abhijitnarvekar:
Hi,

Given that existing applications should function as usual, we need to
support both - text and bytes.

Out of my stupidity :), it occurred to me that we can add one more column
to the table. Lets call it bytesdata of type bytes. And then, we just
change the sessions.py to first check a key in the database configuration.
If it is set to 'bytes', the bytesdata column gets updated. For backward
compatibility, the defaults should refer to the Text type column.  One way
to keep the design database agnostic.

Lastly, forgive me if i m wrong. I am just a beginner in Python.

iredmail:
The question is: does this patch fix the issue for you?

abhijitnarvekar:
I have not had an occasion to try it. I would attempt to as soon
as possible.

PR #649, opened 2020-06-24
Fixed the issue with insert returning query

### LINKED ISSUE #648: query("insert... returning") fails to commit
performing db.query("INSERT into sample_table (value) VALUES ('a')" )works as anticipated: data is stored and committed.

performing db.query("INSERT into sample_table (valid) VALUES ('a') RETURNING idx") fails to commit the insert. Data is added to database, idx is indeed returned, but the insert is not committed, so data is not available.

Problem is due to class DB, query() change with introduction self.create_result_set()

Compare web.py 038 db.py line 661:

    if db_cursor.description:
        ...
        out.list = ...
    else:
        out = db_cursor.rowcount
    if not self.ctx.transactions:
        self.ctx.commit()
    return out

0.51:

    if db_cursor.description:
       return self.create_result_set(db_cursor)  # <---- HERE
    else:
       out = db_cursor.rowcount
    if not self.ctx.transactions:
        self.ctx.commit()
    return out

INSERT...RETURNING sets db_cursor.description, which causes 0.51 to perform `return self.create_result_set(db_cursor)` which means `self.ctx.commit()` is never called.

It appears (with limited testing) a possible solution may be as simple as replacing:

    return self.create_result_set(db_cursor)

with:

    out = self.create_result_set(db_cursor)

and allowing it to flow through to `return out`.

anandology:
Closes #648

### DIFF
--- a/tests/test_db.py
+++ b/tests/test_db.py
@@ -220,6 +220,17 @@ class PostgresTest2(DBTest):
     dbname = "postgres"
     driver = "psycopg2"
 
+    def setUp(self):
+        super().setUp()
+        self.db.query("DROP TABLE IF EXISTS post")
+        self.db.query(
+            "create table post (id serial primary key, title text, body text)"
+        )
+
+    def tearDown(self):
+        self.db.query("DROP TABLE IF EXISTS post")
+        super().tearDown()
+
     def test_limit_with_unsafe_value(self):
         db = self.db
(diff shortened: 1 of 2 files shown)


anandology:
These black issues are really frustrating.

anandology:
Fixed the black errors.

PR #668, opened 2020-08-06
Define a repr() for templates to ease debugging

cclauss:
When debugging, templates appear as `<web.template.Template instance at 0x4036ab8>` which does not aid the developer.

This PR proposes `Template(text="Template text", filename="burndown_chart.html")` instead.

### DIFF
--- a/web/template.py
+++ b/web/template.py
@@ -942,6 +942,22 @@ class Template(BaseTemplate):
             builtins=builtins,
         )
 
+    def __repr__(self):
+        """
+        >>> Template('Template text', 'burndown_chart.html')
+        Template(text="Template text", filename="burndown_chart.html")
+
+        >>> Template('This is too much text with /t & /n characters', 'index.html')
+        Template(text="This is too much text with /t & /n chara", filename="index.html")
+        """
+        try:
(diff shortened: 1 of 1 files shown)


anandology:
I would suggest to just keep the filename and not contents of the file in repr.

cclauss:
[_For many types, this function makes an attempt to return a string that would yield an object with the same value when passed to eval()_](https://docs.python.org/3/library/functions.html#repr)

Unfortunately, [in the current api](https://github.com/webpy/webpy/blob/2ed9ee2d82ef68d063acdba26bb763f15841e5e8/web/template.py#L914-L917) the `text` parameter comes first and is optional and the filename parameter comes second and is optional.  Both suggestions would not pass the `eval()` recommendation above.

cclauss:
Given that the repr() doc entry suggests two alternatives, what about:
* `<web.template.Template instance filename="burndown_chart.html">`

anandology:
> Given that the repr() doc entry suggests two alternatives, what about:
> 
>     * `<web.template.Template instance filename="burndown_chart.html">`

How about just `<Template burndown_chart.html>`

Issue #670, opened 2020-08-09
ResourceWarning: unclosed file <_io.TextIOWrapper


stroeder:
with-statement should be used when calling open() to avoid this warning (when running on Python 3.8):

/usr/lib/python3.8/site-packages/web/template.py:1090: ResourceWarning: unclosed file <_io.TextIOWrapper name='/opt/ae-dir/etc/ae-dir-pwd/templates/en/requestpw_action.html' mode='r' encoding='utf-8'>
   open(path, encoding="utf-8").read(), filename=path, **self._keywords
 ResourceWarning: Enable tracemalloc to get the object allocation traceback

iredmail:
This PR should be closed by https://github.com/webpy/webpy/commit/76c5c7d23000b49d87318745b4bc9a9676f7e143

PR #671, opened 2020-08-09
avoid ResourceWarning by using with-statement when calling open() #670

stroeder:


web/template.py
```diff
@@ -1093,8 +1093,10 @@ def _load_template(self, name):
                 path, cache=self._cache is not None, base=self._base, **self._keywords
             )
         elif kind == "file":
+            with open(path, encoding="utf-8") as tmpl_file:
+                tmpl_str = tmpl_file.read()
             return Template(
-                open(path, encoding="utf-8").read(), filename=path, **self._keywords
+                tmpl_str, filename=path, **self._keywords
```
cclauss, line 1099:
```suggestion
            with open(path, encoding="utf-8") as tmpl_file:
                return Template(tmpl_file.read(), filename=path, **self._keywords)
```

stroeder: Do you really want to place the return-statement in the body of the with-statement?

cclauss: Yes.  This is no problem.  The context manager will close properly even with a return statement.

stroeder: Ok. Applied your suggestion.

web/template.py
```diff
@@ -1093,9 +1093,10 @@ def _load_template(self, name):
                 path, cache=self._cache is not None, base=self._base, **self._keywords
             )
         elif kind == "file":
-            return Template(
-                open(path, encoding="utf-8").read(), filename=path, **self._keywords
-            )
+            with open(path, encoding="utf-8") as tmpl_file:
+                return Template(
+                    tmpl_file.read(), filename=path, **self._keywords
+                )
```
cclauss, line 1099:
```suggestion
                return Template(tmpl_file.read(), filename=path, **self._keywords)
```
To placate that pesky psf/black formatter.

iredmail:
@cclauss Feel free to merge this PR if it's all good.

iredmail:
@cclauss don't forgot to update ChangeLog file. :)

cclauss:
The contributor should update the changelog, not the merger.

iredmail:
We cannot wait for contributor to do everything, so if something is missing and we have some spare time, we'd better do it ourselves, this will help move on a lot faster.

cclauss:
Also, this change is not worthy of a changelog entry as it has no impact on end users.

iredmail:
It fixes a bug, so it's worth a line in ChangeLog.txt. :)

PR #652, opened 2020-07-06
Make sure session DiskStore `cleanup()` correctly returns when session directory was removed.

iredmail:
I recently experienced an issue with app using DiskStore as session store. Seems client didn't visit website for long enough to  trigger the session timeout, also the session directory was removed, when this client visit the website, web.py still tries to check the info of its seession id, since the session directory was removed, this raises exception `OSError` or `FileNotFoundError`.

This PR should fix this issue by checking session directory existence before running `os.path.listdir()`, unittest included.

PR #672, opened 2020-08-19
Handle both directory and file in session cleanup.

iredmail:
fixes #652

### DIFF
--- a/web/session.py
+++ b/web/session.py
@@ -309,15 +309,18 @@ class DiskStore(Store):
             os.remove(path)
 
     def cleanup(self, timeout):
-        if not os.path.isdir(self.root):
-            return
-
         now = time.time()
-        for f in os.listdir(self.root):
-            path = self._get_path(f)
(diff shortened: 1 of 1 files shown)


web/session.py
```diff
@@ -309,15 +309,18 @@ def __delitem__(self, key):
...
-        for f in os.listdir(self.root):
-            path = self._get_path(f)
-            atime = os.stat(path).st_atime
+
+        if os.path.isdir(self.root):
+            for f in os.listdir(self.root):
+                path = self._get_path(f)
+                atime = os.stat(path).st_atime
+                if now - atime > timeout:
+                    shutil.rmtree(path)
```
ryankrumins, line 319:
```suggestion
                    if os.path.isdir(path):
                        shutil.rmtree(path)
                    else:
                        os.remove(path)
```

iredmail: @ryankrumins You're right. Thanks.

PR #677, opened 2020-09-21
Revert an incorrect cookie modification

iredmail:
Commit https://github.com/webpy/webpy/commit/651ea426de61e71ce0ab8a304dbefd29d4611b62#diff-8a8747d48229930871bbbe6e4bd8767eL152 incorrectly modified the cookie state after session killed. It causes browser cookie not cleaned up and web browser session may behave incorrectly.

I experienced the error and fixed it by reverted old commit, tested locally and on production server, both work fine now.


iredmail:
Included in #684.

PR #681, opened 2020-10-13
Handle both session file and directory.

### LINKED ISSUE #680: NotADirectory Error
When a session has expired and web.py creates a new session. the code that attempts to cleanup (i.e., delete) the old session file generates a NotADirectory error.  The cause of the error is the use of shutil.rmtree(path) to remove a file.  It only works if the argument is a directory.  To correct this, change the line to os.remove(path).

Note: I am not (yet) set up to work properly with git and do a pull request.  Since this change is trivial, I hope that you will accept a simple snippet of code showing the change.  Also, I am new to this code base so it is possible that I am missing something in how the code operates, so please confirm that this is the correct change.  Looks good to me and is working for me.  But what do I know anyway.

The change is in sessions.py

248 class Diskstore(Store):
<snip>
310    def cleanup(self, timeout):
311        if not os.path.isdir(self.root):
312            return
313
314        now = time.time()
315        for f in os.listdir(self.root):
316            path = self._get_path(f)
317            atime = os.stat(path).st_atime
318            if now - atime > timeout:
319                shutil.rmtree(path)  ---> Change this line to os.remove(path)

iredmail:
fixes #680

### DIFF
--- a/web/session.py
+++ b/web/session.py
@@ -316,7 +316,10 @@ class DiskStore(Store):
             path = self._get_path(f)
             atime = os.stat(path).st_atime
             if now - atime > timeout:
-                shutil.rmtree(path)
+                if os.path.isdir(path):
+                    shutil.rmtree(path)
+                else:
+                    os.remove(path)
 
 
 class DBStore(Store):


cclauss (commented):
RSLGTM

iredmail:
No idea why GitHub doesn't allow me to merge this PR with approval from one reviewer. :(

cclauss (approved):
It is a ___good___ idea to require at least one positive review.

PR #684, opened 2020-10-21
Few session fixes

### LINKED ISSUE #655: Cannot set session cookie to be a "Session" cookie
0.41 ignored session 'timeout', when setting session cookie & always set the cookie to expire with the browser session.

For some reason 0.50 sets the cookie to have a duration (always) and to set that duration to the value of session 'timeout'.

The problem is there is now no way to return to previous behavior and set a browser Session cookie. web.config.session_parameters['timeout'] must be numeric and must be > 0, as it is _always_ used as the value for session cookie timeout.

If I set session_timeout to 0, login will always fail (because session is too old).
If I set session_timeout to 10, a user who is idle for ten seconds will have their session killed.
If I set session_timeout to 2 days, a user who is idle for two days will have their session killed (I can deal with that), BUT, because the cookie isn't a Session Cookie, it also means if the browser is closed AND A DIFFERENT USER ACCESSES COMPUTER they're logged in as the previous user.

This does not implement a useful "remember-me" feature, though it appears that was the hope.

Needed: some way to force browser Session Cookie.

### LINKED ISSUE #683: can't empty session storage list
This has been working flawlessly in the past version but since I upgraded to new version 0.61 I can't set a session storage value to None or 0 once it has had another value other than that.

Pretty bad since I use it to store shopping cart items, now I need to kill session if I want to empty the cart.

iredmail:
Few session fixes:

* Fixed: can not reset session data to same value as initialized. fixes #683
* Fixed: can not set session expire time. fixes #655
* Fixed: not export session store `MemoryStore`.

### DIFF
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -3,6 +3,8 @@
 ## 2020-XX-XX 0.62
 
 * Fixed: can not reset session data to same value as initialized. #683
+* Fixed: can not set session expire time. #655
+* Fixed: not export session store `MemoryStore`.
 
 ## 2020-06-23 0.61
 
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -1,5 +1,9 @@
 # web.py changelog
 
+## 2020-XX-XX 0.62
+
+* Fixed: can not reset session data to same value as initialized. #683
+
 ## 2020-06-23 0.61
 
 * setup.py: Add python_requires='>=3.5' #662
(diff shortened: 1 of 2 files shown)


iredmail:
Please help review. @cclauss @anandology

iredmail:
Dear @cclauss @anandology,

Please help review.

cclauss (approved):
RSLGTM

PR #685, opened 2020-10-30
Detect key before `del`.

iredmail:
I personally experienced `KeyError: 'ip'` few times, so might be a good idea to fix it with a `if` condition.

### DIFF
--- a/web/session.py
+++ b/web/session.py
@@ -146,7 +146,8 @@ class Session(object):
     def _save(self):
         current_values = dict(self._data)
         del current_values["session_id"]
-        del current_values["ip"]
+        if "ip" in current_values:
+            del current_values["ip"]
 
         if not self.get("_killed"):
             self._setcookie(self.session_id)
--- a/web/session.py
+++ b/web/session.py
@@ -145,9 +145,8 @@ class Session(object):
 
     def _save(self):
         current_values = dict(self._data)
-        del current_values["session_id"]
-        if "ip" in current_values:
-            del current_values["ip"]
+        current_values.pop("session_id", None)
+        current_values.pop("ip", None)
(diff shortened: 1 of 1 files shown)


anandology:
Looking for change of IP is probably a bad idea in itself and that feature in it self can be dropped IMHO.

iredmail:
hi @anandology 

Original code tries to delete key `ip` no matter it exists or not, but it failed few times for me. So i simply add a line to check existence before delete.

I didn't change what the original code do.

cclauss:
`dict.pop("ip", None)` expresses the same idea more compactly.

iredmail:
> `dict.pop("ip", None)` expresses the same idea more compactly.

Thanks for this tip. Updated. :)

iredmail:
Please help review. :)

cclauss:
I believe that @anandology is advocating that we remove `. _validate_ip()` and all code that manipulates `ip` in this class.

iredmail:
I understand. This PR is no harm but fixes one issue. If the best solution is removing some code/feature, then please do it. :)

iredmail:
@cclauss @anandology 
Please help review and merge.

iredmail:
Hi @anandology,

If you have some time, please either merge this PR, or "drop" the feature as mentioned in your first reply in this thread.

ohye3166:
> Hi @anandology,
> 
> If you have some time, please either merge this PR, or "drop" the feature as mentioned in your first reply in this thread.

hello，i meet the same problem， i use your patch and fix it. i don't know why you close it, is there exist a better patch to fix it?

iredmail:
Because no one help review the code, so ...

PR #701, opened 2021-01-06
TEMPLATE_BUILTIN_NAMES: Add Python builtin round()

cclauss:
https://docs.python.org/3/library/functions.html#round useful for internetarchive/openlibrary#4370


anandology (approved):
LGTM!

Issue #710, opened 2021-03-21
DBUtils >= v2.0 changes 'import DBUtils' --> 'import dbutils'


cclauss:
* https://pypi.org/project/DBUtils
* https://webwareforpython.github.io/DBUtils/changelog.html

In DBUtils < v2.0 must use the syntax `import DBUtils`
In DBUtils >= v2.0 must use the syntax `import dbutils`

Where does this repo have references to `DBUtils`:
https://github.com/webpy/webpy/search?q=dbutils

iredmail:
Feel free to change it. :)

cclauss:
@pycarus Which version of https://pypi.org/project/DBUtils are you using?

pycarus:
@cclauss Python 3.8.5 with DB Utils 3.0.1

PR #711, opened 2021-05-15
test: pin dbutils to >= 2.0.0

dhruvmanila:
dbutils 2.0.0 includes breaking changes where the import is done through lowercase letters: `import dbutils` instead of `import DBUtils`

Ref: https://webwareforpython.github.io/DBUtils/changelog.html

Fixes: #710

### DIFF
--- a/tests/test_db.py
+++ b/tests/test_db.py
@@ -147,12 +147,12 @@ class DBTest(unittest.TestCase):
     def testPooling(self):
         # can't test pooling if DBUtils is not installed
         try:
-            import DBUtils  # noqa
+            import dbutils  # noqa
         except ImportError:
             return
         db = setup_database(self.dbname, pooling=True)
         try:
-            self.assertEqual(db.ctx.db.__class__.__module__, "DBUtils.PooledDB")
+            self.assertEqual(db.ctx.db.__class__.__module__, "dbutils.PooledDB")
             db.select("person", limit=1)
         finally:
             db.ctx.db.close()
--- a/web/db.py
+++ b/web/db.py
@@ -656,7 +656,7 @@ class DB:
         self.supports_multiple_insert = False
 
         try:
-            import DBUtils  # noqa, flake8 F401
+            import dbutils  # noqa, flake8 F401
 
             # enable pooling if DBUtils module is available.
             self.has_pooling = True
@@ -712,7 +712,7 @@ class DB:
 
     def _connect_with_pooling(self, keywords):
         def get_pooled_db():
-            from DBUtils import PooledDB
+            from dbutils import PooledDB
 
             # In DBUtils 0.9.3, `dbapi` argument is renamed as `creator`
             # see Bug#122112
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,5 +1,5 @@
 cheroot>=6.0.0
-DBUtils
+DBUtils>=2.0.0
 pytest>=5.4.1
 
 # DB drivers.


dhruvmanila:
I've taken the approach of pinning the dependency as that is only used for testing purposes. If you think otherwise, I will be happy to update it to handle it like so:
```py
try:
    import dbutils
except ImportError:
    import DBUtils as dbutils
except:
    return
```

tests/test_db.py
```diff
@@ -147,12 +147,12 @@ def testNestedTransactions(self):
     def testPooling(self):
         # can't test pooling if DBUtils is not installed
         try:
-            import DBUtils  # noqa
+            import dbutils  # noqa
```
cclauss, line 150: Please be specific about which error(s) we want flake8 to ignore as discussed at
https://flake8.pycqa.org/en/3.9.2/user/violations.html?highlight=noqa#in-line-ignoring-errors

web/db.py
```diff
@@ -656,7 +656,7 @@ def __init__(self, db_module, keywords):
         self.supports_multiple_insert = False
 
         try:
-            import DBUtils  # noqa, flake8 F401
+            import dbutils  # noqa, flake8 F401
```
cclauss, line 659: Please use the proper flake8 ignore syntax
https://flake8.pycqa.org/en/3.9.2/user/violations.html?highlight=noqa#in-line-ignoring-errors

cclauss (changes requested):

cclauss:
Please do
```python
try:
    import dbutils
except ImportError:
    import DBUtils as dbutils
except ImportError:  # No bare except please!
    return
```
https://realpython.com/the-most-diabolical-python-antipattern/

dhruvmanila:
Alright, now that Travis is happy do you like this approach or should I use the above mentioned one? Also, I'm not sure if we can chain exceptions of the same error.

I know this is possible:
```py
try
    ...
except ImportError:
    ...
except IndexError:
    ...
```

But I'm not sure about this, although we can try if that is the preferred way:
```py
try
    ...
except ImportError:
    ...
except ImportError:  # same error as above
    ...
```

web/db.py
```diff
@@ -719,7 +719,7 @@ def get_pooled_db():
             # In DBUtils 0.9.3, `dbapi` argument is renamed as `creator`
             # see Bug#122112
 
-            if PooledDB.version.split(".") < "0.9.3".split("."):
+            if PooledDB.__version__.split(".") < "0.9.3".split("."):
```
cclauss:
```suggestion
            if PooledDB.__version__.split(".") < (0, 9, 3):
```

cclauss:
```python
try:
    ...
except ImportError:
    try:
        ...
    except ImportError:
        ...
```

web/db.py
```diff
@@ -656,7 +656,7 @@ def __init__(self, db_module, keywords):
         self.supports_multiple_insert = False
 
         try:
-            import DBUtils  # noqa, flake8 F401
+            import dbutils  # noqa: F401
 
             # enable pooling if DBUtils module is available.
```
cclauss, line 661: Let’s just simplify…
```python
def get_pooled_db():
            from dbutils import pooled_db as PooledDB
            return PooledDB.PooledDB(creator=self.db_module, **keywords)
```

cclauss: Also, given that the `ast` module has be around since [Python 2.6](https://docs.python.org/2.6/library/ast.html), can you please delete lines 21-25 and then put `import ast` near the top of the file?

dhruvmanila: I don't quite understand regarding your first suggestion. Do you want to remove optional pooling? So, `dbutils` is always necessary? Or, do you just want to get the functionality into a separate function? I suspect it's the later, but just want to confirm.

PR #720, opened 2021-12-03
utils.py: Make `iters` a real tuple, not a hacked tuple

cclauss:
We no longer need to support Python 2.3 so remove the `_hack()` which was meant to support a Python before `set()` and `frozenset()`.

All [uses of `iters`](https://github.com/webpy/webpy/search?q=iters) are for use as the second parameter to `isinstance()` which must be a tuple, not a list.

### DIFF
--- a/web/session.py
+++ b/web/session.py
@@ -233,7 +233,7 @@ class Store:
         return encodebytes(pickled)
 
     def decode(self, session_data):
-        """decodes the data to get back the session dict """
+        """decodes the data to get back the session dict"""
         if isinstance(session_data, str):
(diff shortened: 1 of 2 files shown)


anandology:
I think the reason for using `_hack` was to generate documentation for iters. The script to generate the docs look at `__doc__` and there wasn't any easy way to get docs for variables, so the `_hack_ is used. 

I think it is okay to let go of that legacy, but we should make sure that is not ignored in the docs.

cclauss:
Now I understand.  Let’s leave it as is.

PR #726, opened 2022-03-27
Set daemon attribute instead of using setDaemon method that was deprecated in Python 3.10.

tirkarthi:
The camelcase methods were deprecated in Python 3.10 in https://github.com/python/cpython/pull/25174


cclauss (approved):
Nice!  Thanks

PR #729, opened 2022-05-19
chore: remove unnecessary files in generated python package

aisk:
Current build will add unnecessary files, like mypy cache files, pyc files, `.DS_store` files in tests / docs folder, which will be installed on users' machine.

And the `tools` folder will be a namespace package, which will be installed on users' `site-packages`, and makes a 'tools' module which can be imported.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,5 @@
 include LICENSE.txt
 include requirements.txt
 include test_requirements.txt
-recursive-include docs *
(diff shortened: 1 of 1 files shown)


aisk:
Current release file from [pypi.io](https://pypi.org/project/web.py/#files) containes json files as mypy cache:

web.py package content

iredmail (approved):

cclauss (approved):

PR #730, opened 2022-05-19
refactor: run pyupgrade on exists codes

aisk:
The command is `pyupgrade --keep-percent-format --py3-plus`, for we have python 3.5 and upper version support, and I think percent format is not deprecated in near future.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,5 +40,6 @@ setup(
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
     ],
 )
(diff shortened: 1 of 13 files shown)


cclauss:
In general, I try to upgrade to [f-strings](https://www.scivision.dev/python-f-string-speed) as quickly as possible.

cclauss:
Py36 is sunset too.  I would like to drop **both** and add Py310 if @iredmail agrees.
* https://devguide.python.org/devcycle/#end-of-life-branches
* https://devguide.python.org/#status-of-python-branches

iredmail (commented):
LGTM

cclauss (approved):
LGTM.  Thanks for doing this.

PR #733, opened 2022-06-28
Update README.md

cxtruong70:
Made minor edits to improve readability by removing adverbs and hedging statements like "manually" and "... intended."

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ To install it, please run:
 python3 -m pip install web.py==0.62
 ```
 
-If you are using Python 2.7, then please use web.py version 0.51
+If you are still using Python 2.7, then please use web.py version 0.51
 which is our last release that supports Python 2. 
 ```
 # For Python 2.7
--- a/README.md
+++ b/README.md
@@ -12,15 +12,15 @@ To install it, please run:
 python3 -m pip install web.py==0.62
 ```
 
-If you are still using Python 2.7, then please use web.py version 0.51
-which is intended to be our last release that supports Python 2. 
+If you are using Python 2.7, then please use web.py version 0.51
+which is our last release that supports Python 2. 
 ```
 # For Python 2.7
 python2 -m pip install web.py==0.51
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,15 +12,15 @@ To install it, please run:
 python3 -m pip install web.py==0.62
 ```
 
-If you are still using Python 2.7, then please use web.py version 0.51
-which is intended to be our last release that supports Python 2. 
+If you are using Python 2.7, then please use web.py version 0.51
```
cclauss, line 15:
```suggestion
If you are still using Python 2.7, then please use web.py version 0.51
```
The use of `still` was quite deliberate and is even more appropriate today then when it was first written.

README.md
```diff
@@ -12,15 +12,15 @@ To install it, please run:
...
+which is our last release that supports Python 2. 
 ```
 # For Python 2.7
 python2 -m pip install web.py==0.51
 ```
 
 You can also download it from [GitHub Releases](https://github.com/webpy/webpy/releases)
```
cclauss, line 22: Now that pip contains a real dependency resolver, I no longer see any advantage in recommending this method of installation.

cclauss (changes requested):

README.md
cxtruong70: Good point. We should remove it entirely as the user should never run into dependencies issues related to this.

cxtruong70 (commented):
I have removed the recommendation for Python 2.7 as discussed.

iredmail (approved):

cclauss:
@iredmail Do you understand why tests are failing?

iredmail:
Seems caused by MySQL default charset `utf8mb4`.
FYI https://stackoverflow.com/questions/43437490/pdo-construct-server-sent-charset-255-unknown-to-the-client-please-rep

cclauss:
https://dev.mysql.com/doc/relnotes/mysql/8.0/en/

cclauss (approved):
Thanks for doing this.

PR #737, opened 2022-08-06
Fix session.py to allow suppression of the cookie

bobintetley:
There are often endpoints/targets in your web.py app where you do not want to use sessions and most importantly, you do not want a Set-Cookie header sent with the response. The Set-Cookie header invalidates your Cache-Control directives when you are caching your responses with many CDNs. This PR allows you to do session.no_cookie=True in your handler to suppress the session cookie when you do not want it.


web/session.py
```diff
@@ -147,7 +148,13 @@ def _save(self):
         current_values = dict(self._data)
         del current_values["session_id"]
         del current_values["ip"]
-
+         # Stops the cookie being set for this response
```
cclauss, line 151:
```suggestion
        # Stops the cookie being set for this response
```

iredmail:
Instead of using `no_XXX = False`, i prefer names like `use_XXX = True`. :)

bobintetley:
yep, it could certainly be send_cookie = True with the two booleans reversed. I mean, it's only a 2 line patch so whatever you prefer as maintainers!

cclauss:
Yes, please.  Also please run the code thru psf/black so the tests pass.

bobintetley:
I'm sorry, I don't use psf/black and I have no idea why this is failing as the github error does not offer any indication why. I can't see anything wrong with the 3 lines of code in my patch with regards formatting.

cclauss (dismissed):

cclauss (approved):

PR #743, opened 2022-10-12
pre-commit: Add pyupgrade --py36-plus

cclauss:
Builds on #742 

https://pypi.org/project/pyupgrade

### DIFF
--- a/web/application.py
+++ b/web/application.py
@@ -157,7 +157,7 @@ class application:
         host="0.0.0.0:8080",
         headers=None,
         https=False,
-        **kw
+        **kw,
     ):
         """Makes request to this application for the specified path and method.
         Response will be a storage object with data, status and headers.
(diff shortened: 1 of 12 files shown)


web/application.py
```diff
@@ -410,9 +407,7 @@ def gaerun(self, *middleware):
             minor = version[1]
 
             if major != 2:
-                raise EnvironmentError(
-                    "Google App Engine only supports python 2.5 and 2.7"
-                )
+                raise OSError("Google App Engine only supports python 2.5 and 2.7")
```
iredmail, line 410: Question: Google App Engine (GAE) supports Python 2 + 3, should we improve webpy to support GAE with Python3, or we drop GAE completely?

iredmail (dismissed):

web/application.py
cclauss: Let's do that in a separate PR.

iredmail (approved):

Issue #757, opened 2023-03-01
UnicodeError while handling a bug


criptocoko:
I'm getting the following error while sending an email when a bug occurs:
ERROR HAPPENS ...
...
..
During handling of the above exception, another exception occurred:  

File "/usr/local/lib/python3.10/dist-packages/web/application.py", line 278, in process
    return p(lambda: process(processors))
  File "/usr/local/lib/python3.10/dist-packages/web/application.py", line 686, in processor
    return handler()
  File "/usr/local/lib/python3.10/dist-packages/web/application.py", line 278, in <lambda>
    return p(lambda: process(processors))
  File "/usr/local/lib/python3.10/dist-packages/web/application.py", line 287, in process
    raise self.internalerror()
  File "/usr/local/lib/python3.10/dist-packages/web/debugerror.py", line 354, in emailerrors_internal
    sendmail(
  File "/usr/local/lib/python3.10/dist-packages/web/utils.py", line 1449, in sendmail
    mail.send()
  File "/usr/local/lib/python3.10/dist-packages/web/utils.py", line 1544, in send
    self.send_with_smtp(message_text)
  File "/usr/local/lib/python3.10/dist-packages/web/utils.py", line 1592, in send_with_smtp
    smtpserver.sendmail(self.from_address, self.recipients, message_text)
  File "/usr/lib/python3.10/smtplib.py", line 875, in sendmail
    msg = _fix_eols(msg).encode('ascii')
UnicodeEncodeError: 'ascii' codec can't encode character '\\u2019' in position 12430: ordinal not in range(128)

If i add message_text.enconde('ascii", errors="ignore") the problem is fixed but i don't want to patch webpy code.... is there a way to fix this?

iredmail:
Please show us code snippet to reproduce this issue.

criptocoko:
Sure, here is an example of web.debugerror.emailerrors_internal() that generates a non unicode char

`
import web

web.config.smtp_server="postfix" # valid smtp
web.config.smtp_port=587 

web.utils.sendmail( "your buggy site<root@example.com>", 
"the bugfixer<bugfixer@example.com>", 
"bug:bla bla", "message", 
attachments=[dict(filename="bug.html", content=web.utils.safestr('  bla ’ bla '))]) ## example of attachemnt with a non unicode char  '\u2019' 
`

criptocoko:
[...]

PR #758, opened 2023-03-01
fix attachment unicodeError

criptocoko:
This is a fix to handle emails with attachments that have characters that are not in range(128), in this case we encode them in base64.

### DIFF
--- a/web/utils.py
+++ b/web/utils.py
@@ -1516,7 +1516,12 @@ class _EmailMessage:
 
         if not content_type.startswith("text/"):
             encoders.encode_base64(msg)
-
+        else:
+            try:
+                msg = msg.encode("ascii")
+            except Exception as e:
+                encoders.encode_base64(msg)
+                
         self.message.attach(msg)
 
     def prepare_message(self):


web/utils.py
```diff
@@ -1514,9 +1514,9 @@ def attach(self, filename, content, content_type=None):
         msg.add_header("Content-Type", content_type)
         msg.add_header("Content-Disposition", "attachment", filename=filename)
 
-        if not content_type.startswith("text/"):
+        if not content_type.startswith("text/") or not msg.isascii():
             encoders.encode_base64(msg)
-
+
```
iredmail, line 1519: Please remove space/tab in this line.

iredmail (changes requested):

iredmail:
Hi @criptocoko,

As requested in #757, could you please show us code snippet to reproduce the issue?

iredmail:
Hi @cclauss, could you help check the `black --check .` issue in GitHub Actions result?
I ran the latest `black` version locally but it's all fine, not sure why it failed in GitHub Actions.

```
$ black --check .
All done! ✨ 🍰 ✨
29 files would be left unchanged.
```

iredmail:
By the way, this change looks good to me after checked source code in py `encoders.encode_base64()`.

iredmail (dismissed):

PR #761, opened 2023-03-02
Use ternary operator instead of if-else-block

cclauss:
[`ruff --select=SIM108 --fix .`](https://beta.ruff.rs/docs/rules/#flake8-simplify-sim)


iredmail:
Code is shorter but the readability seems worse to me. :(

anandology:
I feel these changes are making code hard to read.

PR #762, opened 2023-03-04
[FIXED] # UnicodeError while handling a bug #757

Z4nzu:
This error occurs when there is an attempt to encode a Unicode character that is not in the ASCII range (i.e., outside the range of 0 to 127) using the ASCII codec. The character in question in your case is '\u2019', which is the right single quotation mark.

Encode the message using a Unicode encoding that supports all characters, such as UTF-8. You can do this by changing the encoding parameter of the encode() method to 'utf-8'

### DIFF
--- a/web/debugerror.py
+++ b/web/debugerror.py
@@ -355,7 +355,7 @@ def emailerrors(to_address, olderror, from_address=None):
             "your buggy site <%s>" % from_address,
             "the bugfixer <%s>" % to_address,
             "bug: %(error_name)s: %(error_value)s (%(path)s)" % locals(),
-            message,
+            message.encode('utf-8'),
             attachments=[dict(filename="bug.html", content=safestr(djangoerror()))],
         )
         return error


Z4nzu:
Hii @cclauss, Please Review PR

web/debugerror.py
```diff
@@ -355,7 +355,7 @@ def emailerrors_internal():
             "your buggy site <%s>" % from_address,
             "the bugfixer <%s>" % to_address,
             "bug: %(error_name)s: %(error_value)s (%(path)s)" % locals(),
-            message,
+            message.encode('utf-8'),
```
cclauss: We run https://pypi.org/project/black
```suggestion
            message.encode("utf-8"),
```

web/debugerror.py
```diff
@@ -355,7 +355,7 @@ def emailerrors_internal():
             "your buggy site <%s>" % from_address,
             "the bugfixer <%s>" % to_address,
             "bug: %(error_name)s: %(error_value)s (%(path)s)" % locals(),
-            message,
+            message.encode("utf-8"),
```
cclauss, line 358:
```suggestion
            message.encode("utf-8") if isinstance(message, str) else message,
```
A bit safer...
```python
>>> for message in ("string", b"bytes"):
...     message.encode("utf-8")
...
b'string'
Traceback (most recent call last):  <-- We do not want exceptions to be raised!!
  File "<stdin>", line 2, in <module>
AttributeError: 'bytes' object has no attribute 'encode'. Did you mean: 'decode'?
>>> for message in ("string", b"bytes"):
...     message.encode("utf-8") if isinstance(message, str) else message
...
b'string'
b'bytes'
```

cclauss (approved):
Nice!  Thanks for doing this.

cclauss:
@iredmail @criptocoko Your reviews, please?

cclauss (approved):

criptocoko:
Does this solve the scenario i presented?? because i think the problem is only for the attachments, not the message. The message is already handled here: https://github.com/webpy/webpy/blob/0b67153411870604197349159a167a3a1133f705/web/utils.py#L1483. Correct me if im wrong.

I attached the 2 scenarios('non ascii on the attach and the non ascii on the message') where the first one fails and second doesn't.
web.py issue

cclauss:
This is _exactly_ why tests are a great idea.

cclauss:
Please resolve the git conflict.

PR #764, opened 2023-04-20
Some minor bug fixes / updates

marksmayo:
Replacing strings with f-strings
Fixing bare exceptions
Cleaning whitespace/brackets
conditions on empty lists
reorder exceptions where one is an ancestor exception
remove useless return
update getName() to name ala python 3.11 requirements

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -2,7 +2,7 @@ from web import utils
 
 
 def test_group():
-    assert list(utils.group([], 2)) == []
+    assert not list(utils.group([], 2))
     assert list(utils.group([1, 2, 3, 4, 5, 6, 7, 8, 9], 3)) == [
         [1, 2, 3],
         [4, 5, 6],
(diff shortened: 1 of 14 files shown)


cclauss:
* [ ] Please edit `pyproject.toml` and add `PLC1901` to the `ignore` list.
* [ ] Do the `except:` changes allow us to add `BLE` to the select list?
* [ ] Do these changes allow us to remove anything from the `ignore` list?

https://github.com/webpy/webpy/blob/0b67153411870604197349159a167a3a1133f705/pyproject.toml#L62-L77

cclauss:
% `pip3 install pipx`
% `pipx run ruff --select=UP031 --fix .`

marksmayo:
I think there may still be a ruff failiure in the next run, but it's not appearing locally for me, so just working out the difference.

cclauss:
You can[ install pre-commit locally](https://pre-commit.com/#install) and then do:
pre-commit install
pre-commit autoupdate
pre-commit run --all-files

tests/test_utils.py
```diff
@@ -2,7 +2,7 @@
 
 
 def test_group():
-    assert list(utils.group([], 2)) == []
+    assert not list(utils.group([], 2))
```
cclauss, line 5: Please revert ALL falsey values would pass this test including None, False, 0, 0.0, {}, set(), (,).

We are asserting that the result is an empty list, not any of these other values.

Repeat below.  https://docs.astral.sh/ruff/rules/compare-to-empty-string/#known-problems

PR #773, opened 2023-09-20
Remove cgi.FieldStorage (based on #769)

scottbarnes:
I took the changes from #769 and modified them as little as possible to get Open Library working with it. See internetarchive/openlibrary#8313.

This still needs more testing.

@tfmorris
@cclauss 
@mekarpeles

PR #778, opened 2023-10-27
Fix storify and multipart

### LINKED ISSUE #776: UnicodeDecodeError in 0.70
Alas, it appears I missed something in #769, specifically with respect to `utils.py`.

With 0.70 and the Open Library code base, the following occurs when `web.input()` in called when trying to read an attached file, such as a book cover sent via POST as a multipart attachment:
```python
  File "/home/openlibrary/.local/lib/python3.11/site-packages/web/application.py", line 276, in process
    return self.handle()
           ^^^^^^^^^^^^^
  File "/home/openlibrary/.local/lib/python3.11/site-packages/web/application.py", line 267, in handle
    return self._delegate(fn, self.fvars, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/openlibrary/.local/lib/python3.11/site-packages/web/application.py", line 509, in _delegate
    return handle_class(cls)
           ^^^^^^^^^^^^^^^^^
  File "/home/openlibrary/.local/lib/python3.11/site-packages/web/application.py", line 487, in handle_class
    return tocall(*args)
           ^^^^^^^^^^^^^
  File "/openlibrary/openlibrary/coverstore/code.py", line 164, in POST
    i = web.input(
        ^^^^^^^^^^
  File "/home/openlibrary/.local/lib/python3.11/site-packages/web/webapi.py", line 474, in input
    return storify(out, *requireds, **defaults)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/openlibrary/.local/lib/python3.11/site-packages/web/utils.py", line 184, in storify
    value = getvalue(value)
            ^^^^^^^^^^^^^^^
  File "/home/openlibrary/.local/lib/python3.11/site-packages/web/utils.py", line 168, in getvalue
    if hasattr(x, "file") and hasattr(x, "value"):
                              ^^^^^^^^^^^^^^^^^^^
  File "/home/openlibrary/.local/lib/python3.11/site-packages/multipart.py", line 425, in value
    return self.raw.decode(self.charset)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

This is likely because the content of `.file` has changed in `multipart`, and `webpy` shouldn't be trying to decode it as unicode when it's a binary file.

Updating `getvalue()` in `utils.py` to the following fixes covers on Open Library, but I still need to compare the output to the `cgi.FieldStorage` equivalent:
```python
def getvalue(x):
    if hasattr(x, 'file') and hasattr(x, 'raw'):                                                                                          
        # There's an attached file; return the content directly.
        return x.file.read()
    elif hasattr(x, 'value'):                
        return unicodify(x.value)                                  
    else:                   
        return unicodify(x)
```

How did I miss this? I think because I did not rebuild the containers with `--no-cache` I had a cached copy of the `covers` container, and because no tests exercised the code in `utils.py`, there was no warning about `cgi` being deprecated, so I falsely thought `cgi` was no longer in use. However, when doing final testing and rebuilding the containers with `--no-cache` I found I was mistaken.

I will look more at this tomorrow and submit hopefully submit a PR.

Note, 0.70 is not yet on PyPi, so people who are using `pip install webpy` are not affected by this currently, as they will still get 0.62: https://pypi.org/project/web.py/#history.

@cclauss

scottbarnes:
Closes #776 

This surely needs more testing. I tried to make the different commits meaningful and targeted on one thing. Specifically, 6318bfa73a086953d0a746c231e89674e4a45617 should be the one with the substantive changes (of which there are few -- it's mostly test).

I also made this responsive to some of the feedback from @tfmorris (thank you!) on #773. I have not yet make it responsive to all -- specifically the reference to `"_fieldstorage"`, which I agree should likely go.

I have been sitting on this PR for a while, though, and I want to get a draft of it in so I can perhaps get some feedback on whether this approach seems good.

I did test it with Open Library, and it works there, for whatever that is worth.

@cclauss

### DIFF
--- a/web/utils.py
+++ b/web/utils.py
@@ -169,10 +169,6 @@ def storify(mapping, *requireds, **defaults):
             return x.file.read()
         else:
             return unicodify(getattr(x, "value", x))
-        # elif hasattr(x, "value"):
-        #     return unicodify(x.value)
-        # else:
-        #     return unicodify(x)
 
     stor = Storage()
     for key in requireds + tuple(mapping.keys()):
--- a/web/utils.py
+++ b/web/utils.py
@@ -165,12 +165,14 @@ def storify(mapping, *requireds, **defaults):
             return s
 
     def getvalue(x):
-        if hasattr(x, "file") and hasattr(x, "value"):
-            return x.value
-        elif hasattr(x, "value"):
-            return unicodify(x.value)
+        if hasattr(x, "file") and hasattr(x, "raw"):
+            return x.file.read()
         else:
-            return unicodify(x)
+            return unicodify(getattr(x, "value", x))
+        # elif hasattr(x, "value"):
+        #     return unicodify(x.value)
+        # else:
+        #     return unicodify(x)
 
     stor = Storage()
     for key in requireds + tuple(mapping.keys()):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,3 +1,7 @@
+from io import BytesIO
+
+from multipart import parse_form_data
+
 from web import utils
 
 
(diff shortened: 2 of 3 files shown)


web/utils.py
```diff
@@ -165,12 +165,14 @@ def unicodify(s):
...
-            return unicodify(x.value)
+        if hasattr(x, "file") and hasattr(x, "raw"):
+            return x.file.read()
         else:
-            return unicodify(x)
+            return unicodify(getattr(x, "value", x))
+        # elif hasattr(x, "value"):
+        #     return unicodify(x.value)
+        # else:
+        #     return unicodify(x)
```
mekarpeles, line 175:
```suggestion
```

rm unused code paths

scottbarnes: Whoops. Thanks for catching that.

mekarpeles:
One more courtesy ping, I think @scottbarnes's change here is necessary to unbreak the latest build.

@cclauss do you have a moment to 👀

tfmorris (commented):
The code is definitely improved and I didn't see anything needing fixing in my cursory review, but the main thing that needs to be verified is test coverage and I don't have time to verify the upload file test case right now.

If it works for the OpenLibrary use case, I think we can circle back and clean up any other loose ends later.

I haven't reviewed the corresponding OL change, but if code modifications were needed, it would be good to abstract them into an "How to update your application" note here for any other consumers.

mekarpeles:
Thank you @tfmorris for investing the time. Will circle back after verifying on the open Library side that minimal changes were needed and provide whatever migration documentation is required

cclauss (approved):

Issue #781, opened 2023-11-30
Please add nginx unit as "production" platform!


jn0:
Please fix `wsgi.py` (around https://github.com/webpy/webpy/blob/master/web/wsgi.py#L75) to read as

```python
    if (
        "SERVER_SOFTWARE" in os.environ
        or "PHP_FCGI_CHILDREN" in os.environ
        or "fcgi" in argv
        or "fastcgi" in argv
        or "mod_wsgi" in argv
        or "unit" in argv  # PLEASE ADD THIS FIX
    ):
        return False
```

jn0:
> Did you actually test webpy app with Nginx Unit?

Yes, we run it in production.

igorlidin:
Let me confirm the proposed enhancement, we also (successfully) use webpy with Nginx Unit.

cclauss:
Please add TESTS.

jn0:
> Please add TESTS.

I cannot realize what is to be tested here?

If one would run the WSGI app in Nginx Unit, then he'll face the word "unit" passed in sys.argv (I've checked by logging argv).
So, the test is to run an application under Unit.

Do you need a sample app (it's a common web.py based WSGI) or what?

iredmail:
@cclauss i think we can simply add that line in webpy code.

PR #782, opened 2023-12-19
wsgi.py: Add nginx unit as a production platform

cclauss:
Fixes: #781
* #781

@jn0 Your review, please.


iredmail:
I prefer the multiple `or xxx in argv` edition, it's more readable. :)

cclauss:
`str.split()` is difficult to understand.

PR #802, opened 2025-02-04
Remove __import__ from TEMPLATE_BUILTIN_NAMES to prevent RCE

### LINKED ISSUE #801: __import__ can be used inside a template to bypass security restrictions
At line 789 of template.py, the TEMPLATE_BUILTIN_NAMES list includes the __import__ statement. This statement can be leveraged to execute arbitrary commands inside a template, bypassing the security restrictions.

The following payload is an example of how this can be achieved:
`${__import__('os').system('<command_here>')}`

gehaxelt:
Fixes #801

### DIFF
--- a/web/template.py
+++ b/web/template.py
@@ -786,7 +786,7 @@ TEMPLATE_BUILTIN_NAMES = [
     "True",
     "False",
     "None",
-    "__import__",  # some c-libraries like datetime requires __import__ to present in the namespace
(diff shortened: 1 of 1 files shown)


anandology:
This may break backward compatibility as libraries like `datetime` will stop working.

gehaxelt:
@anandology I think we can close this one, since #803 fixes the issue and was merged.

PR #803, opened 2025-02-06
Fix: disable accessing special variables from template

### LINKED ISSUE #801: __import__ can be used inside a template to bypass security restrictions

anandology:
Before this fix, it was possible to import modules from templates, which opens the possibility of RCE. This applies only when templates from untrusted users is allowed in the application. While this is not a common scenario, given the design goals of web.py templating system to be sandboxable, this is a serious flaw.

This commit fixes that issue by rejecting access to special variables with names starting with `__` at the compile time.

Fixes #801.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,5 +88,5 @@ lint.pylint.max-branches = 17 # default is 12
 lint.pylint.max-returns = 8 # default is 6
 
 [tool.codespell]
-ignore-words-list = "asend,gae,notin"
+ignore-words-list = "asend,gae,notin,requireds"
 skip = [ "./.*" ]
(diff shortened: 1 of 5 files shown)


anandology:
@cclauss codespell is failing for `utils.py` and `webapi.py`. How did this go unnoticed until now? I haven't made any changes to those file.

anandology:
@scottbarnes @cclauss @gehaxelt could you please review this PR?

cclauss:
Codespell has occasional updates that add new misspellings that may cause new failures.

mekarpeles (approved):
Changes look good to me.

Issue #806, opened 2025-04-21
How to deal with CVE-2025-3818?


kongyi182:
We find a new vulnerability in webpy 0.70: CVE-2025-3818, how to deal with CVE-2025-3818?

PR #807, opened 2025-04-21
Address CVE-2025-3818

mekarpeles:
Addresses #806

The SQL vulnerability in the previous implementation lies in the lack of validation or sanitization of the seqname parameter before it is incorporated into the SQL query. Without proper validation, an attacker could exploit this by injecting malicious SQL code into the seqname parameter, leading to a SQL Injection vulnerability.

Solution ensures that seqname only contains valid characters for SQL sequence names (letters, numbers, and underscores), and that it starts with a valid letter or underscore. This eliminates the possibility of injecting malicious SQL code.

The SQLQuery object is assumed to safely handle the sequence name as a parameter, properly escaping it or using parameterized queries to prevent SQL injection.

### DIFF
--- a/web/db.py
+++ b/web/db.py
@@ -1205,7 +1205,7 @@ class PostgresDB(DB):
     def get_sequence_query(self, seqname):
         import re
         # Ensure the sequence name is valid
-        if not re.match(r'^[a-zA-Z_][a-zA-Z0-9_]*$', seqname):
+        if not re.match(r'^[a-zA-Z_][a-zA-Z0-9_$]*$', seqname):
             raise ValueError(f"Invalid sequence name: {seqname}")
         return SQLQuery("; SELECT currval(%s)", seqname)
 
--- a/web/db.py
+++ b/web/db.py
@@ -1198,9 +1198,17 @@ class PostgresDB(DB):
                 seqname = None
 
         if seqname:
-            query += "; SELECT currval('%s')" % seqname
+            query += self.get_sequence_query(seqname)
 
         return query
+
+    def get_sequence_query(self, seqname):
+        import re
+        # Ensure the sequence name is valid
+        if not re.match(r'^[a-zA-Z_][a-zA-Z0-9_]*$', seqname):
+            raise ValueError(f"Invalid sequence name: {seqname}")
+        return SQLQuery("; SELECT currval(%s)", seqname)
+
 
     def _get_all_sequences(self):
         """Query postgres to find names of all sequences used in this database."""


web/db.py
```diff
@@ -1198,10 +1198,18 @@ def _process_insert_query(self, query, tablename, seqname):
...
         if seqname:
-            query += "; SELECT currval('%s')" % seqname
+            query += self.get_sequence_query(seqname)
 
         return query
 
+    def get_sequence_query(self, seqname):
+        import re
+        # Ensure the sequence name is valid
+        if not re.match(r'^[a-zA-Z_][a-zA-Z0-9_]*$', seqname):
```
scottbarnes, line 1208:
```suggestion
        if not re.match(r'^[a-zA-Z_][a-zA-Z0-9_$]*$', seqname):
```
I am just basing this suggestion on PosgreSQL documentation section [4.4.1 Identifiers and Key Words](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS), which says in relevant part:
> SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($). Note that dollar signs are not allowed in identifiers according to the letter of the SQL standard, so their use might render applications less portable. The SQL standard will not define a key word that contains digits or starts or ends with an underscore, so identifiers of this form are safe against possible conflict with future extensions of the standard.

However, I take the point that the SQL standard does not allow `$`, so maybe it's not worth including?

scottbarnes (approved):
LGTM.

web/db.py
```diff
@@ -1198,10 +1198,18 @@ def _process_insert_query(self, query, tablename, seqname):
...
         if seqname:
-            query += "; SELECT currval('%s')" % seqname
+            query += self.get_sequence_query(seqname)
 
         return query
 
+    def get_sequence_query(self, seqname):
+        import re
+        # Ensure the sequence name is valid
+        if not re.match(r'^[a-zA-Z_][a-zA-Z0-9_$]*$', seqname):
```
anandology, line 1208: It may be a good idea to compile the regex outside the function or use `utils.re_compile`.

anandology (commented):
Everything else looks good to me. Thanks for addressing this issue.

web/db.py
```diff
@@ -1198,10 +1198,18 @@ def _process_insert_query(self, query, tablename, seqname):
...
+            raise ValueError(f"Invalid sequence name: {seqname}")
+        return SQLQuery("; SELECT currval(%s)", seqname)
```
cclauss, line 1210: Why favor the negative?
```suggestion
        if re.match(r'^[a-zA-Z_][a-zA-Z0-9_$]*$', seqname):
            return SQLQuery("; SELECT currval(%s)", seqname)
        raise ValueError(f"Invalid sequence name: {seqname}")
```

anandology:
It reads better to have error checks first and the main body later. It is
hard to read if the most important code of the function is tucked inside an
if condition.

eamanu:
Hi! I'd like to do a ping regarding to this issue. Looks good and it would very useful if you can release it :-)

mekarpeles:
I think I'll have to run some lint fixes; will do that momentarily.

PR #808, opened 2025-05-07
pre-commit run --all-files, fix SQLQuery

cclauss:


web/db.py
```diff
@@ -1204,12 +1204,12 @@ def _process_insert_query(self, query, tablename, seqname):
 
     def get_sequence_query(self, seqname):
         import re
+
         # Ensure the sequence name is valid
-        if not re.match(r'^[a-zA-Z_][a-zA-Z0-9_$]*$', seqname):
+        if not re.match(r"^[a-zA-Z_][a-zA-Z0-9_$]*$", seqname):
             raise ValueError(f"Invalid sequence name: {seqname}")
         return SQLQuery("; SELECT currval(%s)", seqname)
```
mekarpeles, line 1211: I think my original PR should have done this instead based on the constructor in https://github.com/webpy/webpy/blob/master/web/db.py#L132-L145:

```suggestion
        return SQLQuery(f"SELECT currval('{seqname}')")
```

PR #827, opened 2026-06-09
pyproject.toml: rework license information

yegorich:
Don't use license info in the classifiers section define both license and license file via the new options.

This resolves the following issues:

```
*********************************************************************** 
Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

This deprecation is overdue, please update your project and remove deprecated calls to avoid build errors in the future.

See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details. 
***********************************************************************
```

```
*********************************************************************** 
Please consider removing the following classifiers in favor of a SPDX license expression:

        License :: Public Domain

See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details. 
***********************************************************************
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,11 +20,11 @@ repos:
       # args: [--fix, --exit-non-zero-on-fix]
 
 -   repo: https://github.com/abravalheri/validate-pyproject
-    rev: v0.23
+    rev: v0.25
     hooks:
     - id: validate-pyproject
 
 -   repo: https://github.com/tox-dev/pyproject-fmt
-    rev: v2.5.0
+    rev: v2.23.0
     hooks:
     - id: pyproject-fmt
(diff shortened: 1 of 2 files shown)


yegorich:
There is another consideration: there is no "Public domain" license as an SPDX definition. So I had to create a custom tag. The closest definition would be [CC0-1.0](https://spdx.org/licenses/CC0-1.0.html).

yegorich:
Do we need to bump setuptools to 77.0.0?

cclauss:
### `pre-commit`: upgrade `validate-pyproject` and `pyproject-fmt`
Let's make sure that the `pyproject.toml` tools validate the proposed changes.
* https://github.com/abravalheri/validate-pyproject/releases
* https://github.com/tox-dev/pyproject-fmt/releases

cclauss (approved):
@anandology @mekarpeles Can you please comment on whether we should choose the [custom](https://packaging.python.org/en/latest/glossary/#term-License-Expression) `LicenseRef-Public-Domain-1.0` string conforming to the SPDX specification or the SPDX identifier [`CC0-1.0`](https://spdx.org/licenses/CC0-1.0.html)?

PR #830, opened 2026-09-09
fix broken doctest examples in net.py

remnantblackbox:
I've tried to run net.py and received these failures:

```
3 items had failures:
   1 of   1 in __main__.htmlquote
   1 of   1 in __main__.htmlunquote
   2 of   3 in __main__.websafe
***Test Failed*** 4 failures.
``` 

I've figured out that these failures were caused by the u string prefix in the expected outputs in htmlquote, htmlunquote and websafe, as shown in this error example:

```
Failed example:
    websafe(None)
Expected:
    u''
Got:
    ''
```

So i removed them because in python3 they are no longer necessary. Since i touched only docstrings, no logic was changed and when I reran net.py I received no output, so the doctests passed successfully.

### DIFF
--- a/web/net.py
+++ b/web/net.py
@@ -221,7 +221,7 @@ def htmlquote(text):
     Encodes `text` for raw use in HTML.
 
         >>> htmlquote(u"<'&\">")
-        u'&lt;&#39;&amp;&quot;&gt;'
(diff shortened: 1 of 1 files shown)


cclauss (approved):
This prefix is not needed on Python 3 because ALL `str` are Unicode and `bytes != str`.

